PR #9, opened 2022-03-21
Clear engine use

russellmorley:
Clear Engine uses these changes in derived classes to include additional processing to TextSegments before training and to intercept versification. Making these virtual enhance the re-usability of the framework.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/9)


russellmorley:
Damien, in this commit, https://github.com/sillsdev/machine/pull/9/commits/bc8f8d513e6a55f5a8d40e4d03166b4338278f7a , I pulled out a change I needed to compile the WPF test app that leaked into another commit. I didn't have time to determine why the change was needed to compile but suspect it may be a nuget dependency version.

PR #16, opened 2022-07-27
Allowing for keyless unit testing.

javamonkey79:


tests/SIL.Machine.WebApi.SpecFlowTests/StepDefinitions/MachineAPIStepDefinitions.cs
```diff
@@ -18,22 +18,6 @@ public sealed class MachineApiStepDefinitions
         public MachineApiStepDefinitions()
         {
             this.client = new WebApiClient(MACHINE_API_TEST_URL, true);
-            SetAccessTokenFromEnvironment();
```
johnml1135, line 21: We should keep this code but make it a warning instead of an error.   We may also use this to test a QA environment and want to verify that the authentication is working.

johnml1135 (changes requested):
the lack of API client and secret should be a warning, not an error.

tests/SIL.Machine.WebApi.SpecFlowTests/StepDefinitions/MachineAPIStepDefinitions.cs
javamonkey79: @johnml1135 done. It is not clear to me the best way to log a warning in C# also, it looks like the ternary operator needs to throw an exception in order to work. Thus, I've gone with the try...catch approach. LMK if it's not what you hoped for.

johnml1135:
Sorry again, the throwing was only if it returned a null.  The throws should be removed and the signals just checked for null.

javamonkey79:
> Sorry again, the throwing was only if it returned a null. The throws should be removed and the signals just checked for null.

All good @johnml1135 ! Please check again.

johnml1135:
I merged it into my gherkin testing branch.

PR #36, opened 2023-05-12
refractor deployment to serval and other updates

johnml1135:
* The k8s deployment moved to serval (and removed from here)
  * Update github actions to create numbered docker images
  * Update VSCode launcher to attach to docker containers from serval deployment
* Fixes/updates to the code:
  * Allow for nmt machine.py docker image to be specified through environment
  * Allow for saving/not saving images through  environment (may be better to be passed through grpc and the api)
  * Throttle SMT building status reports to serval to max 1/second
  * nmt pretranslate on all specified textid - even if a translation already exists
  * Fix ClearML authentication
  * Fix clearml project delete
  * Fix mongo item delete filter
  * Fix "need specific directories" issue for docker containers
  * Fix suggest command, specifying parameter name of "cancellationToken"

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)


src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs
```diff
@@ -258,7 +258,6 @@ async IAsyncEnumerable<Pretranslation> ProcessRowsAsync()
                     if (
                         (corpus.PretranslateAll || corpus.PretranslateTextIds.Contains(row.TextId))
                         && row.SourceSegment.Count > 0
-                        && row.TargetSegment.Count == 0
```
ddaspit, line 261: This check should not be removed. We purposefully don't generate pretranslations for verses that already have a translation.

src/SIL.Machine.AspNetCore/Services/BuildProgress.cs
```diff
@@ -6,6 +6,10 @@ public class BuildProgress : IProgress<ProgressStatus>
     private readonly string _buildId;
     private ProgressStatus _prevStatus;
 
+    private DateTime lastReportTime = DateTime.Now;
```
ddaspit, line 9: Private fields should be prefixed with an underscore.

src/SIL.Machine.AspNetCore/Services/BuildProgress.cs
```diff
@@ -6,6 +6,10 @@ public class BuildProgress : IProgress<ProgressStatus>
...
+
+    private float throttleTimeSeconds = 1;
```
ddaspit, line 11: This should be a constant.

src/SIL.Machine.AspNetCore/Configuration/ClearMLNmtEngineOptions.cs
```diff
@@ -10,5 +10,7 @@ public class ClearMLNmtEngineOptions
     public string SecretKey { get; set; } = "";
     public TimeSpan BuildPollingTimeout { get; set; } = TimeSpan.FromSeconds(2);
     public int MaxStep { get; set; } = 500_000;
+    public bool SaveModel { get; set; } = true;
```
ddaspit, line 13: I am removing the code that saves a model in machine.py, so this won't be necessary.

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
```diff
@@ -12,10 +13,19 @@ public class ClearMLService : IClearMLService
             Converters = { new CustomEnumConverterFactory(JsonNamingPolicy) }
         };
 
-    public ClearMLService(HttpClient httpClient, IOptionsMonitor<ClearMLNmtEngineOptions> options)
+    private string authorizationToken = "";
```
ddaspit, line 16: These fields should be prefixed with underscores.

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
```diff
@@ -204,16 +216,42 @@ public async Task<bool> StopTaskAsync(string id, CancellationToken cancellationT
...
         {
             Content = new StringContent(body.ToJsonString(), Encoding.UTF8, "application/json")
         };
+        request.Headers.Add("Authorization", $"Bearer {authorizationToken}");
+        HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
+        string result = await response.Content.ReadAsStringAsync();
+        return (JsonObject?)JsonNode.Parse(result);
+    }
+
+    private async Task AuthorizeAsync(CancellationToken cancellationToken = default)
```
ddaspit, line 230: Does this access token expire? If so, we will need to check if it is expired and request a new access token. Also, this makes the service stateful. It is possible for multiple threads to call this method concurrently causing race conditions. Locking should be used to avoid these issues.

src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs
```diff
@@ -31,7 +31,7 @@ public IDistributedReaderWriterLock Create(string id)
 
     public async Task<bool> DeleteAsync(string id, CancellationToken cancellationToken = default)
     {
-        RWLock? rwLock = await _locks.DeleteAsync(id, cancellationToken);
+        RWLock? rwLock = await _locks.DeleteAsync(e => (string)e.Id == id, cancellationToken);
```
ddaspit, line 34: Why did you change this to a lambda expression?

src/SIL.Machine.AspNetCore/Services/SharedFileService.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace SIL.Machine.AspNetCore.Services;
+﻿using System.IO;
```
ddaspit, line 1: You can add this using to the `Usings.cs` file.

ddaspit (changes requested):
Just a couple of suggested changes.

src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs
johnml1135: What is the rationale behind that?  I can think of a few examples when people might want to:
* To manually compare against what is already written
* If it was drafted but we wanted to see what a new draft would look like with a new model

Basically, this behavior could be replicated through just not uploading the specific texts for training - yet since we are specifically calling out which texts to pretranslate, it makes sense to me to translate all of it.  By contrast, we could also add another flag called "forcePretranslate/pretranslateExistingText" or some other term that would have a description.  Whatever we decide now, we should at least update the documentation (or track the updating of it).

src/SIL.Machine.AspNetCore/Services/BuildProgress.cs
johnml1135: Done.

src/SIL.Machine.AspNetCore/Configuration/ClearMLNmtEngineOptions.cs
johnml1135: Removed.

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
johnml1135: Could we change ClearMLService to have a Scoped rather than Singleton lifetime?

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
johnml1135: Done.

src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs
johnml1135: It didn't work with the id just as a string - it threw an error.  I looked into it and this was the clearest resolution.  It works.

src/SIL.Machine.AspNetCore/Services/SharedFileService.cs
johnml1135: I think it was added automatically when making an update - I was able to remove it and it complied fine.

src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs
johnml1135: Revert it for now - as per Damien "It will generate about the same text already".  Update Swagger documentation to specify this behavior.

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
johnml1135: Check if creating a new access token will invalidate existing ones.  Check if they have lifetimes.

johnml1135: Singleton should be acceptable as long as the access token lifetime and race conditions are taken into account.

johnml1135:
The ClearML auth login expires after 30 days (https://clear.ml/docs/latest/docs/references/api/#request-format).  I updated the code to refresh the token every hour.  It should be thread safe.

src/SIL.Machine.AspNetCore/Services/ClearMLService.cs
johnml1135: Changed from getting a new auth token each time a request is made to periodically refreshing the authentication token (every hour).  That removes any race condition, accounts for the lifetime of the token (max 30 days - it would have broken after a month and we would not know why - good catch) and should be ok if the token is being refreshed when there is a call, as the "old" tokens are not invalidated until their timer runs out.

johnml1135 (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36#-:-NVf2H_ZAzfDjM7nhdHl:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLService.cs` line 230 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193375195:-NVfD9Mz9VznZlvX4Ei9:bpgob1k) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/ClearMLService.cs#L230)):*

This is a good solution. It should be implemented as a [hosted service](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-7.0&tabs=visual-studio). Also, the engine server and the job server will both be communicating with ClearML and need the access token. When one of the servers retrieves a token, it will invalidate the other's token. To solve this, we should store the token in the database, so both servers can use the same token. The engine server can run the hosted service to keep it up-to-date.
</blockquote></details>

I just tested it out - we can have multiple bearer tokens at the same time with the same creds.  Each one has it's own lifecycle and is not invalidated if a new one is made.  I see no reason to add the complexity of having a central service manage the creds (assuming it is added complexity).

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLService.cs` line 230 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193375195:-NVgu-TUDsRkEE6JWvgN:b-kptsrh) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/ClearMLService.cs#L230)):*

I just tested it out - we can have multiple bearer tokens at the same time with the same creds.  Each one has it's own lifecycle and is not invalidated if a new one is made.  I see no reason to add the complexity of having a central service manage the creds (assuming it is added complexity).
</blockquote></details>

That is good. We don't need to worry about saving the token to the database. This code still needs to be converted to a hosted service.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs` line 34 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193379735:-NVj9tQd0EWIVT_BhmhG:b-6sioeu) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs#L34)):*

The error is thrown when trying to delete a ClearML engine only - and here is the message:

Exception has occurred: CLR/MongoDB.Driver.Linq.ExpressionNotSupportedException
Exception thrown: 'MongoDB.Driver.Linq.ExpressionNotSupportedException' in MongoDB.Driver.dll: 'Expression not supported: Convert(e, IEntity).Id in (Convert(e, IEntity).Id == "64662c9ee00320874d9c4847") because field "_id" is not represented as a string.'
   at MongoDB.Driver.Linq.Linq3Implementation.Translators.ExpressionToFilterTranslators.MethodTranslators.StringExpressionToRegexFilterTranslator.TranslateField(TranslationContext context, Expression expression, Expression fieldExpression)
</blockquote></details>

I think this is the issue: https://jira.mongodb.org/browse/CSHARP-4522.  Specifically, in DataAccess 0.3.0, you are using MongoDriver v. 2.19.0 which has the stated bugs.  By updating to 2.19.1 (and publishing a new DataAccess), the issue should be resolved.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs` line 34 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193379735:-NVjqjN_EeGxPKRGMKvj:b1sb9kj) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs#L34)):*

I think this is the issue: https://jira.mongodb.org/browse/CSHARP-4522.  Specifically, in DataAccess 0.3.0, you are using MongoDriver v. 2.19.0 which has the stated bugs.  By updating to 2.19.1 (and publishing a new DataAccess), the issue should be resolved.
</blockquote></details>

Reverted - we will need to publish the new DataAccess to fully resolve this for NMT.

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)* status: all files reviewed, 10 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 109 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmIeo235yKI5pEzbhX:-NVmIeo235yKI5pEzbhY:btwbzyz) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L109)):*
> ```Smalltalk
>             p => p.GetRequiredService<ClearMLAuthenticationService>()
>         );
>         builder.Services.AddSingleton<IHostedService>(p => p.GetRequiredService<ClearMLAuthenticationService>());
> ```

You should use `AddHostedService`.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 8 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLYTm5tNn9slAgw5-:-NVmLYTm5tNn9slAgw50:b82dl7i) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L8)):*
> ```Smalltalk
>     private readonly IOptionsMonitor<ClearMLNmtEngineOptions> _options;
>     private readonly ILogger<ClearMLAuthenticationService> _logger;
>     private int _refreshPeriod = 3600; // 1 hour
> ```

This should be a constant.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLaGd7cc3Hw5zTUE0:-NVmLaGd7cc3Hw5zTUE1:bblgdc3) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L10)):*
> ```Smalltalk
>     private int _refreshPeriod = 3600; // 1 hour
>     private PeriodicTimer? _periodicTimer = null;
>     private CancellationTokenSource _ctSource = new CancellationTokenSource();
> ```

This should be `readonly`.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 11 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmHyX90yniH3Xuo9gQ:-NVmHyX90yniH3Xuo9gR:bqksxkk) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L11)):*
> ```Smalltalk
>     private PeriodicTimer? _periodicTimer = null;
>     private CancellationTokenSource _ctSource = new CancellationTokenSource();
>     private int executionCount = 0;
> ```

What is the intent of this field?
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 34 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmNFfH6juqJ5UEyxGC:-NVmNFfH6juqJ5UEyxGD:b-ieft53) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L34)):*
> ```Smalltalk
>         _logger.LogInformation("ClearML Authentication Token Refresh service running.");
> 
>         InitRefreshAuthorizationTimer();
> ```

I would wrap this in a `Task.Run` call to make it clear that it is running in a separate thread, i.e.

```
Task.Run(InitRefreshAuthorizationTimer);
```
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmHY3DE_KMtEgpUVVY:-NVmHY3DE_KMtEgpUVVZ:b1a2dp0) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L39)):*
> ```Smalltalk
>     }
> 
>     private async void InitRefreshAuthorizationTimer()
> ```

This should return a `Task`.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 44 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLevyD9o0T7Yx8Gvl:-NVmLevyD9o0T7Yx8Gvm:b-un3s3q) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L44)):*
> ```Smalltalk
> 
>         // call once during init so that it is populated
>         await AuthorizeAsync(expiration_sec: _refreshPeriod * 2 + 10, cancellationToken: token);
> ```

Why are you doubling the refresh period and adding 10 seconds? You should schedule the refresh before the token expires. Also, you do it again inside the `AuthorizeAsync` method.

The initial `AuthorizeAsync` call should occur inside `StartAsync` so that we ensure that we have a token.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 50 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmL2nI35V-0Y-3yzbE:-NVmL2nI35V-0Y-3yzbF:b-33vb0c) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L50)):*
> ```Smalltalk
>         try
>         {
>             while (await _periodicTimer.WaitForNextTickAsync(cancellationToken: token))
> ```

I don't think the `PeriodicTimer` is necessary. You can just use `await Task.Delay();` in a loop.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 59 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmJV_GDRdxxYwr6ZOi:-NVmJV_GDRdxxYwr6ZOj:b-mixnuc) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L59)):*
> ```Smalltalk
>     }
> 
>     private async Task AuthorizeAsync(int expiration_sec, CancellationToken cancellationToken)
> ```

The convention for variables in C# is to use camel case.
___
*[`src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs` line 34 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193379735:-NVmEsiu2MYXvvQZIEWF:b-ya5y9n) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs#L34)):*

Reverted - we will need to publish the new DataAccess to fully resolve this for NMT.
</blockquote></details>

Machine is already using 2.19.1, so that isn't the issue. I can't reproduce the problem. Maybe the data is corrupted. Can you give me specific steps to reproduce the crash?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)* status: all files reviewed, 11 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 41 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmO0nX4q-xIjN8aJA8:-NVmO0nYDx7kSO1jM25v:b27w8li) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L41)):*
> ```Smalltalk
>     private async void InitRefreshAuthorizationTimer()
>     {
>         var token = _ctSource.Token;
> ```

Please remember to only use `var` when the type is explicit.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 109 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmIeo235yKI5pEzbhX:-NVoClc67BMJFRO5UsAu:bqczc42) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L109)):*

You should use `AddHostedService`.
</blockquote></details>

It won't work, as per the discussions here (https://stackoverflow.com/questions/52036998/how-do-i-get-a-reference-to-an-ihostedservice-via-dependency-injection-in-asp-ne) and here (https://github.com/dotnet/extensions/issues/553).  Specifically, we want it to be a singleton and to be accessible from another class.  I believe this is the proper implementation.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 8 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLYTm5tNn9slAgw5-:-NVoDuCn5edQRx5UZ_f-:bb6un1j) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L8)):*

This should be a constant.
</blockquote></details>

Done

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLaGd7cc3Hw5zTUE0:-NVoE6KhEP4ciTxSvWqR:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L10)):*

This should be `readonly`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 11 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmHyX90yniH3Xuo9gQ:-NVoEGTXCs_nNtCDHUAe:b-q5wqu7) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L11)):*

What is the intent of this field?
</blockquote></details>

Monitoring - to see how many times it was refreshed.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 34 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmNFfH6juqJ5UEyxGC:-NVoFIJp2FYsCwJXkhn5:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L34)):*

I would wrap this in a `Task.Run` call to make it clear that it is running in a separate thread, i.e.

```
Task.Run(InitRefreshAuthorizationTimer);
```
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmHY3DE_KMtEgpUVVY:-NVoFMbXBWhLkuDBh3Vb:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L39)):*

This should return a `Task`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 41 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmO0nX4q-xIjN8aJA8:-NVoFRr27ZGPTqIV7UBN:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L41)):*

Please remember to only use `var` when the type is explicit.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 11 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmHyX90yniH3Xuo9gQ:-NVoHlI19kb6dhgme8ty:b-38y5ek) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L11)):*

Monitoring - to see how many times it was refreshed.
</blockquote></details>

Changed - now it counts consecutive failures - more useful - and actionable.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 41 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmO0nX4q-xIjN8aJA8:-NVoHtrxDdtIpe9RlZsn:bcvasyb) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L41)):*

Done.
</blockquote></details>

renamed variable.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 44 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmLevyD9o0T7Yx8Gvl:-NVoHwZF-37ecX_1nhLb:b-ncom3b) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L44)):*

Why are you doubling the refresh period and adding 10 seconds? You should schedule the refresh before the token expires. Also, you do it again inside the `AuthorizeAsync` method.

The initial `AuthorizeAsync` call should occur inside `StartAsync` so that we ensure that we have a token.

</blockquote></details>

Removed expiration_sec.  I'll just let it be live for 30 days.  No operational reason for artificially shortening the length of the token being valid.  I don't see it as a security concern - just an operational concern if the token does not refresh successfully for some reason.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 50 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmL2nI35V-0Y-3yzbE:-NVoJYAo9XRVACe0qWS6:b6n4jqu) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L50)):*

I don't think the `PeriodicTimer` is necessary. You can just use `await Task.Delay();` in a loop.
</blockquote></details>

Refactored with Task.Delay.  Is that what you were thinking?

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 59 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmJV_GDRdxxYwr6ZOi:-NVoJehT7M8NjvwW2aOV:bn7635j) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L59)):*

The convention for variables in C# is to use camel case.
</blockquote></details>

Sorry - removed variable.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs` line 34 at r1](https://reviewable.io/reviews/sillsdev/machine/36#gh-1193379735:-NVojizgEnDbdGw3ROo1:b-870lxw) ([raw file](https://github.com/sillsdev/machine/blob/2ede0b94424fbb95a0ebf270a2596cc083470d30/src/SIL.Machine.AspNetCore/Services/DistributedReaderWriterLockFactory.cs#L34)):*

Machine is already using 2.19.1, so that isn't the issue. I can't reproduce the problem. Maybe the data is corrupted. Can you give me specific steps to reproduce the crash?
</blockquote></details>

I made the fix and published dataaccess 0.3.1.  It resolves the issue.

ddaspit (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 109 at r3](https://reviewable.io/reviews/sillsdev/machine/36#-NVmIeo235yKI5pEzbhX:-NW0m7XfAUflZG25ncGh:b9zuyyf) ([raw file](https://github.com/sillsdev/machine/blob/d28d006eb1c119c691942688b82740b32610be9e/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L109)):*

It won't work, as per the discussions here (https://stackoverflow.com/questions/52036998/how-do-i-get-a-reference-to-an-ihostedservice-via-dependency-injection-in-asp-ne) and here (https://github.com/dotnet/extensions/issues/553).  Specifically, we want it to be a singleton and to be accessible from another class.  I believe this is the proper implementation.
</blockquote></details>

`AddHostedService` adds the class as a [singleton](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/ServiceCollectionHostedServiceExtensions.cs). The following clearly shows that `ClearMLAuthenticationService` is registered as an API service and a hosted service:

```
builder.Services.AddSingleton<IClearMLAuthenticationService, ClearMLAuthenticationService>();
builder.Services.AddHostedService(p => p.GetRequiredService<IClearMLAuthenticationService>());
```
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs` line 37 at r4](https://reviewable.io/reviews/sillsdev/machine/36#-NW0oiTgADyUAySbD5Ip:-NW0oiTgADyUAySbD5Iq:b-48s4c0) ([raw file](https://github.com/sillsdev/machine/blob/48602380fae5362699769b6b3f616b993223165a/src/SIL.Machine.AspNetCore/Services/ClearMLAuthenticationService.cs#L37)):*
> ```Smalltalk
> 
>         // call once during init to ensure that the auth token is populated
>         AuthorizeAsync(cancellationToken).Wait();
> ```

You should `await` this call, i.e. `await AuthorizeAsync(cancellationToken);`.

ddaspit (approved):
I wanted to get this in, so I went ahead and fixed up the last couple of issues.

> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36#-:-NW0y2uZ7cCeflKHkWd8:b-pge3v5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

ddaspit (commented):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/36)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

Issue #75, opened 2023-09-07
S3 bucket - lockdown when creds are exposed


johnml1135:
If S3 cred get exposed, there is a partial lock-down placed on the account, preventing deletion of data.  This caused our system to crash.  We should not fail a build if we cannot delete the files at the end - log the error, but not fail.  This will prevent a catastrophic downtime for Serval if it happens again.

PR #76, opened 2023-09-07
#75 - don't fail nmt job on delete failing

johnml1135:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/76)

### DIFF
--- a/src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs
+++ b/src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs
@@ -140,7 +140,17 @@ public class ClearMLNmtEngineBuildJob
                 CancellationToken.None
             );
 
-            await _sharedFileService.DeleteAsync($"builds/{buildId}/", CancellationToken.None);
+            try
+            {
+                //Don't fail the whole job if we can't delete the files.
+                await _sharedFileService.DeleteAsync($"builds/{buildId}/", CancellationToken.None);
+            }
+            catch (AmazonS3Exception e)
+            {
+                _logger.LogError(
+                    $"Could not delete build ({buildId}) because of exception {e.GetType().Name}:{e.Message}.  Check S3 bucket key permissions.  Finishing up build anyway."
+                );
+            }
 
             await _engines.UpdateAsync(
                 e => e.EngineId == engineId && e.BuildId == buildId,


Enkidu93 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/76#-:-NdjuPbc12m-6lcyK2iL:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/76)* status: :shipit: complete! all files reviewed, all discussions resolved

ddaspit (commented):
I added a retroactive comment.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/76#-:-Ndk6SDLCTnJejjYp9JH:b-i0i1lm)* status: all files reviewed, 1 unresolved discussion

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs` line 150 at r1](https://reviewable.io/reviews/sillsdev/machine/76#-Ndk6Due0nkkuk0a7Ywb:-Ndk6Due0nkkuk0a7Ywc:b-sc953f) ([raw file](https://github.com/sillsdev/machine/blob/a736a896f4aa010ac3a7be61ef19210431c2c303/src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs#L150)):*
> ```Smalltalk
>             catch (AmazonS3Exception e)
>             {
>                 _logger.LogError(
> ```

You can pass the exception directly to the `LogError` method. This way the stack trace wouldn't be lost. It is a good idea to always pass the exception when logging an error.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs` line 150 at r1](https://reviewable.io/reviews/sillsdev/machine/76#-Ndk6Due0nkkuk0a7Ywb:-Ndk6tVS4W8f_NtuW66e:b-poakls) ([raw file](https://github.com/sillsdev/machine/blob/a736a896f4aa010ac3a7be61ef19210431c2c303/src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs#L150)):*

You can pass the exception directly to the `LogError` method. This way the stack trace wouldn't be lost. It is a good idea to always pass the exception when logging an error.
</blockquote></details>

By the time it gets there, we already know what happened (we have read and write privileges, but not delete).  The error stack trace becomes less valuable.  That is why I chose to not include it.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/76)* status: all files reviewed, 1 unresolved discussion

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs` line 150 at r1](https://reviewable.io/reviews/sillsdev/machine/76#-Ndk6Due0nkkuk0a7Ywb:-Ndk8iMY8E1sr6kdtXv4:bd6offp) ([raw file](https://github.com/sillsdev/machine/blob/a736a896f4aa010ac3a7be61ef19210431c2c303/src/SIL.Machine.AspNetCore/Services/ClearMLNmtEngineBuildJob.cs#L150)):*

By the time it gets there, we already know what happened (we have read and write privileges, but not delete).  The error stack trace becomes less valuable.  That is why I chose to not include it.
</blockquote></details>

You should always use [this](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-loggerextensions-logerror(microsoft-extensions-logging-ilogger-system-exception-system-string-system-object())) overload when logging an exception. It is just good practice.

PR #85, opened 2023-09-15
Fixes S3 bucket unecessary alerts

Enkidu93:
Change interval of health checks to 4 minutes.

S3 pings up to three times before returning Unhealthy.

(Also, removed redundant logging)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)

### DIFF
--- a/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs
+++ b/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs
@@ -101,6 +101,7 @@ public static class IMachineBuilderExtensions
     public static IMachineBuilder AddClearMLService(this IMachineBuilder builder)
     {
         builder.Services.AddSingleton<IClearMLService, ClearMLService>();
+        //Add retry policy; fail after approx. 2 + 4 + 8 = 14 seconds
         builder.Services
             .AddHttpClient<ClearMLService>()
             .AddTransientHttpErrorPolicy(
@@ -110,12 +111,14 @@ public static class IMachineBuilderExtensions
         // workaround register satisfying the interface and as a hosted service.
         builder.Services.AddSingleton<IClearMLAuthenticationService, ClearMLAuthenticationService>();
         builder.Services.AddHostedService(p => p.GetRequiredService<IClearMLAuthenticationService>());
+        //Add retry policy; fail after approx. 2 + 4 + 8 = 14 seconds
         builder.Services
             .AddHttpClient<IClearMLAuthenticationService, ClearMLAuthenticationService>()
             .AddTransientHttpErrorPolicy(
                 b => b.WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)))
             );
 
+        builder.Services.AddSingleton<S3HealthCheck>();
         builder.Services.AddHealthChecks().AddCheck<ClearMLHealthCheck>("ClearML Health Check");
 
         return builder;
--- a/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs
+++ b/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs
@@ -101,17 +101,22 @@ public static class IMachineBuilderExtensions
     public static IMachineBuilder AddClearMLService(this IMachineBuilder builder)
     {
         builder.Services.AddSingleton<IClearMLService, ClearMLService>();
-        builder.Services.AddHealthChecks().AddCheck<ClearMLHealthCheck>("ClearML Health Check");
+        builder.Services
+            .AddHttpClient<ClearMLService>()
+            .AddTransientHttpErrorPolicy(
+                b => b.WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)))
+            );
 
         // workaround register satisfying the interface and as a hosted service.
         builder.Services.AddSingleton<IClearMLAuthenticationService, ClearMLAuthenticationService>();
         builder.Services.AddHostedService(p => p.GetRequiredService<IClearMLAuthenticationService>());
-
         builder.Services
             .AddHttpClient<IClearMLAuthenticationService, ClearMLAuthenticationService>()
             .AddTransientHttpErrorPolicy(
                 b => b.WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)))
             );
+
+        builder.Services.AddHealthChecks().AddCheck<ClearMLHealthCheck>("ClearML Health Check");
 
         return builder;
     }
--- a/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs
+++ b/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs
@@ -5,10 +5,7 @@ public static class IServiceCollectionExtensions
     public static IMachineBuilder AddMachine(this IServiceCollection services, IConfiguration? configuration = null)
     {
         services.AddSingleton<ISharedFileService, SharedFileService>();
-        services.Configure<HealthCheckPublisherOptions>(options =>
-        {
-            options.Period = TimeSpan.FromSeconds(240);
-        });
+        services.AddSingleton<S3HealthCheck>();
         services.AddHealthChecks().AddCheck<S3HealthCheck>("S3 Bucket");
         services.AddScoped<IDistributedReaderWriterLockFactory, DistributedReaderWriterLockFactory>();
         services.AddSingleton<ICorpusService, CorpusService>();
--- a/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs
+++ b/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs
@@ -5,6 +5,10 @@ public static class IServiceCollectionExtensions
     public static IMachineBuilder AddMachine(this IServiceCollection services, IConfiguration? configuration = null)
     {
         services.AddSingleton<ISharedFileService, SharedFileService>();
+        services.Configure<HealthCheckPublisherOptions>(options =>
+        {
+            options.Period = TimeSpan.FromSeconds(240);
+        });
         services.AddHealthChecks().AddCheck<S3HealthCheck>("S3 Bucket");
         services.AddScoped<IDistributedReaderWriterLockFactory, DistributedReaderWriterLockFactory>();
         services.AddSingleton<ICorpusService, CorpusService>();
--- a/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
+++ b/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
@@ -2,8 +2,6 @@ public class ClearMLHealthCheck : IHealthCheck
 {
     private readonly HttpClient _httpClient;
     private readonly IOptionsMonitor<ClearMLNmtEngineOptions> _options;
-    private string _authToken;
-    private readonly AsyncLock _lock;
     private readonly IClearMLAuthenticationService _clearMLAuthenticationService;
 
     public ClearMLHealthCheck(
@@ -14,8 +12,6 @@ public class ClearMLHealthCheck : IHealthCheck
     {
         _httpClient = httpClient;
         _options = options;
-        _authToken = "";
-        _lock = new AsyncLock();
         _clearMLAuthenticationService = clearMLAuthenticationService;
     }
 
@@ -26,9 +22,6 @@ public class ClearMLHealthCheck : IHealthCheck
     {
         try
         {
-            using (await _lock.LockAsync())
-                if (_authToken == "")
-                    _authToken = await _clearMLAuthenticationService.GetAuthTokenAsync(cancellationToken);
             if (!await PingAsync(cancellationToken))
                 return HealthCheckResult.Unhealthy("ClearML is unresponsive");
             if (!await WorkersAreAssignedToQueue(cancellationToken))
@@ -52,7 +45,10 @@ public class ClearMLHealthCheck : IHealthCheck
         {
             Content = new StringContent(body.ToJsonString(), Encoding.UTF8, "application/json")
         };
-        request.Headers.Add("Authorization", $"Bearer {_authToken}");
+        request.Headers.Add(
+            "Authorization",
+            $"Bearer {await _clearMLAuthenticationService.GetAuthTokenAsync(cancellationToken)}"
+        );
         HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
         string result = await response.Content.ReadAsStringAsync(cancellationToken);
         return (JsonObject?)JsonNode.Parse(result);
--- a/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
+++ b/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
@@ -1,10 +1,16 @@
 public class ClearMLHealthCheck : IHealthCheck
 {
-    private readonly ClearMLService _clearMLService;
+    private readonly HttpClient _httpClient;
+    private readonly IOptionsMonitor<ClearMLNmtEngineOptions> _options;
+    private string _authToken;
+    private readonly AsyncLock _lock;
 
-    public ClearMLHealthCheck(ClearMLService clearMLService)
+    public ClearMLHealthCheck(HttpClient httpClient, IOptionsMonitor<ClearMLNmtEngineOptions> options)
     {
-        _clearMLService = clearMLService;
+        _httpClient = httpClient;
+        _options = options;
+        _authToken = "";
+        _lock = new AsyncLock();
     }
 
     public async Task<HealthCheckResult> CheckHealthAsync(
@@ -14,11 +20,12 @@ public class ClearMLHealthCheck : IHealthCheck
     {
         try
         {
-            bool success = await _clearMLService.PingAsync();
-            if (!success)
+            using (await _lock.LockAsync())
+                if (_authToken == "")
+                    _authToken = await GetAuthTokenAsync(cancellationToken);
+            if (!await PingAsync(cancellationToken))
                 return HealthCheckResult.Unhealthy("ClearML is unresponsive");
-            bool workersAvailable = await _clearMLService.WorkersAreAssignedToQueue();
-            if (!workersAvailable)
+            if (!await WorkersAreAssignedToQueue(cancellationToken))
                 return HealthCheckResult.Unhealthy("No ClearML agents are available");
             return HealthCheckResult.Healthy("ClearML is available");
         }
@@ -26,5 +33,64 @@ public class ClearMLHealthCheck : IHealthCheck
         {
             return HealthCheckResult.Unhealthy(exception: e);
         }
+    }
+
+    private async Task<string> GetAuthTokenAsync(CancellationToken cancellationToken)
+    {
+        var request = new HttpRequestMessage(HttpMethod.Post, $"{_options.CurrentValue.ApiServer}/auth.login")
+        {
+            Content = new StringContent("{}", Encoding.UTF8, "application/json")
+        };
+        var authenticationString = $"{_options.CurrentValue.AccessKey}:{_options.CurrentValue.SecretKey}";
+        var base64EncodedAuthenticationString = Convert.ToBase64String(Encoding.ASCII.GetBytes(authenticationString));
+        request.Headers.Add("Authorization", $"Basic {base64EncodedAuthenticationString}");
+        HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
+        string result = await response.Content.ReadAsStringAsync(cancellationToken);
+        return (string)((JsonObject?)JsonNode.Parse(result))?["data"]?["token"]!;
+    }
+
+    private async Task<JsonObject?> CallAsync(
+        string service,
+        string action,
+        JsonNode body,
+        CancellationToken cancellationToken = default
+    )
+    {
+        var request = new HttpRequestMessage(HttpMethod.Post, $"{_options.CurrentValue.ApiServer}/{service}.{action}")
+        {
+            Content = new StringContent(body.ToJsonString(), Encoding.UTF8, "application/json")
+        };
+        request.Headers.Add("Authorization", $"Bearer {_authToken}");
+        HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
+        string result = await response.Content.ReadAsStringAsync(cancellationToken);
+        return (JsonObject?)JsonNode.Parse(result);
+    }
+
+    public async Task<bool> PingAsync(CancellationToken cancellationToken = default)
+    {
+        JsonObject? result = await CallAsync("debug", "ping", new JsonObject(), cancellationToken);
+        return result is not null;
+    }
+
+    public async Task<bool> WorkersAreAssignedToQueue(CancellationToken cancellationToken = default)
+    {
+        JsonObject? result = await CallAsync("workers", "get_all", new JsonObject(), cancellationToken);
+        JsonNode? workers_node = result?["data"]?["workers"];
+        if (workers_node is null)
+            return false;
+        JsonArray workers = (JsonArray)workers_node;
+        foreach (var worker in workers)
+        {
+            JsonNode? queues_node = worker?["queues"];
+            if (queues_node is null)
+                continue;
+            JsonArray queues = (JsonArray)queues_node;
+            foreach (var queue in queues)
+            {
+                if ((string?)queue?["name"] == _options.CurrentValue.Queue)
+                    return true;
+            }
+        }
+        return false;
     }
 }
--- a/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
+++ b/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
@@ -1,8 +1,8 @@
 public class ClearMLHealthCheck : IHealthCheck
 {
-    private readonly IClearMLService _clearMLService;
+    private readonly ClearMLService _clearMLService;
 
-    public ClearMLHealthCheck(IClearMLService clearMLService)
+    public ClearMLHealthCheck(ClearMLService clearMLService)
     {
         _clearMLService = clearMLService;
     }
--- a/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
+++ b/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs
@@ -4,13 +4,19 @@ public class ClearMLHealthCheck : IHealthCheck
     private readonly IOptionsMonitor<ClearMLNmtEngineOptions> _options;
     private string _authToken;
     private readonly AsyncLock _lock;
+    private readonly IClearMLAuthenticationService _clearMLAuthenticationService;
 
-    public ClearMLHealthCheck(HttpClient httpClient, IOptionsMonitor<ClearMLNmtEngineOptions> options)
+    public ClearMLHealthCheck(
+        IClearMLAuthenticationService clearMLAuthenticationService,
+        HttpClient httpClient,
+        IOptionsMonitor<ClearMLNmtEngineOptions> options
+    )
     {
         _httpClient = httpClient;
         _options = options;
         _authToken = "";
         _lock = new AsyncLock();
+        _clearMLAuthenticationService = clearMLAuthenticationService;
     }
 
     public async Task<HealthCheckResult> CheckHealthAsync(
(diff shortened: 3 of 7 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NeP9ByeAxOUMTT5wpr5:b1ec9w) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*
> ```Smalltalk
>     {
>         services.AddSingleton<ISharedFileService, SharedFileService>();
>         services.Configure<HealthCheckPublisherOptions>(options =>
> ```

I don't think we are using `IHealthCheckPublisher` anywhere. Why is this change needed?
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 6 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NePCFJn7yHk2cWGJ5JN:-NePCFJn7yHk2cWGJ5JO:bjkqo7r) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L6)):*
> ```Smalltalk
>     private readonly ILogger _logger;
> 
>     public S3HealthCheck(ISharedFileService sharedFileService, ILoggerFactory loggerFactory)
> ```

I know this wasn't done in this PR, but there is no need to pass the `ILoggerFactory` interface. You can just pass in a `ILogger<S3HealthCheck>` instance directly.
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 19 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9RPN1_qn7inhPr8h:-NeP9RPOCNcwV9r7pdgu:b-8o52z0) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L19)):*
> ```Smalltalk
>         int numConsecutiveFailures = 0;
>         Exception exception = new Exception();
>         while (numConsecutiveFailures < 3)
> ```

The `CheckHealthAsync` method is called whenever the `/health` endpoint is hit. This implementation could cause the endpoint to hang for a long time. Generally, we want fast responses to endpoints. I would suggest that if the check fails, we return a degraded status. If it continues to fail for longer than 90 seconds, then we return unhealthy. We can do this by registering this class as a singleton and storing the time of the first failure as a field. Every time it fails, we check the time to see if it has been longer than 90 seconds. We will need to use a lock when checking/updating the field, since this is a multi-threaded environment.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NePL3619PVs9F2pjXvb:b-rwfgs5) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

I don't think we are using `IHealthCheckPublisher` anywhere. Why is this change needed?
</blockquote></details>

This updates the default healthcheck update interval (as I understand it - I don't entirely understand why/how).
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 6 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NePCFJn7yHk2cWGJ5JN:-NePLG670fsyFtQuAMf-:b-7641dh) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L6)):*

I know this wasn't done in this PR, but there is no need to pass the `ILoggerFactory` interface. You can just pass in a `ILogger<S3HealthCheck>` instance directly.
</blockquote></details>

OK, I can take care of that - thank you. Yeah, I think that was code I wrote early on when I knew even less what I was doing. 
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 19 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9RPN1_qn7inhPr8h:-NePLNsR8mGa-7IJDiYa:b-n4o7sl) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L19)):*

The `CheckHealthAsync` method is called whenever the `/health` endpoint is hit. This implementation could cause the endpoint to hang for a long time. Generally, we want fast responses to endpoints. I would suggest that if the check fails, we return a degraded status. If it continues to fail for longer than 90 seconds, then we return unhealthy. We can do this by registering this class as a singleton and storing the time of the first failure as a field. Every time it fails, we check the time to see if it has been longer than 90 seconds. We will need to use a lock when checking/updating the field, since this is a multi-threaded environment.
</blockquote></details>

Yes, I wanted to do exactly what you were saying but I couldn't find a way to get the fields to persist. But yes, registering as a singleton - that's the way to do it - thank you!

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NePQoX1Cc6avFTbg4H2:b-k8oepl) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

This updates the default healthcheck update interval (as I understand it - I don't entirely understand why/how).
</blockquote></details>

This doesn't have any impact on our health checks. This only affects `IHealthCheckPublisher` implementations, which we aren't using.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NePVkuj8_IVc9rE8MzC:b4v36f4) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

This doesn't have any impact on our health checks. This only affects `IHealthCheckPublisher` implementations, which we aren't using.
</blockquote></details>

 With options.Period set to 4 minutes:

```

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:15:44

...

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:19:44

```

  
With options.Period set to 30 seconds:

```

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:21:58

...

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:22:28

```

 This is talking about gRPC health checks and mentions this solution: [https://learn.microsoft.com/en-us/aspnet/core/grpc/health-checks?view=aspnetcore-7.0#configure-health-checks-execution-interval](https://learn.microsoft.com/en-us/aspnet/core/grpc/health-checks?view=aspnetcore-7.0#configure-health-checks-execution-interval). I tried it, and it works generally. I can't seem to find a way to configure this in a semantically clearer way. Like I said, "I don't understand why/how", but it definitely has an effect. If this isn't the right way to do it, do you know what the right way is?

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NePWZd656rV2ljpYdnD:b-r65qsx) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

With options.Period set to 4 minutes:

```

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:15:44

...

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:19:44

```

  
With options.Period set to 30 seconds:

```

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:21:58

...

machine-engine-cntr | info: S3HealthCheck\[0\]

machine-engine-cntr | 09/15/2023 20:22:28

```

 This is talking about gRPC health checks and mentions this solution: [https://learn.microsoft.com/en-us/aspnet/core/grpc/health-checks?view=aspnetcore-7.0#configure-health-checks-execution-interval](https://learn.microsoft.com/en-us/aspnet/core/grpc/health-checks?view=aspnetcore-7.0#configure-health-checks-execution-interval). I tried it, and it works generally. I can't seem to find a way to configure this in a semantically clearer way. Like I said, "I don't understand why/how", but it definitely has an effect. If this isn't the right way to do it, do you know what the right way is?
</blockquote></details>

(And of course, maybe we don't want to tweak the checking interval - which is fine too - just let me know).

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NePgwwK80hH_eWNln-m:b-b2lyt8) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

(And of course, maybe we don't want to tweak the checking interval - which is fine too - just let me know).
</blockquote></details>

Yes, you are correct. I didn't realize that the gRPC Health Checks service registers a `IHealthCheckPublisher`. The gRPC health check API includes a `Watch` endpoint that can stream health check reports from the server to the client. The publisher is required to support this endpoint. If you fix the S3 health check to not hang, then I don't think we need to change this.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NecTEaHENjOV0e12lOe:bq5e6tx) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

Yes, you are correct. I didn't realize that the gRPC Health Checks service registers a `IHealthCheckPublisher`. The gRPC health check API includes a `Watch` endpoint that can stream health check reports from the server to the client. The publisher is required to support this endpoint. If you fix the S3 health check to not hang, then I don't think we need to change this.
</blockquote></details>

I've realized, Damien, that because of the retry policy on the HttpClient/AmazonS3Client for the ClearML/S3 Healthchecks, the endpoint will hang regardless. What's the best way to handle this? I hate to have the health checks use custom code to interact with AWS/ClearML rather than using our existing service code both for maintainability and consistency, but I'm not sure how else to have this policy one place and not the other. I could generate and close the client on each call and have optional policy parameters in the function calls, but that seems excessive and potentially expensive (?).

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 107 at r2](https://reviewable.io/reviews/sillsdev/machine/85#-Necbn4J5oTjibVU1grW:-Necbn4J5oTjibVU1grX:bw19yhd) ([raw file](https://github.com/sillsdev/machine/blob/c400520d95636cf98a7a61b0737a4435e7944155/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L107)):*
> ```Smalltalk
>             .AddHttpClient<ClearMLService>()
>             .AddTransientHttpErrorPolicy(
>                 b => b.WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)))
> ```

When will this actually error out?  Can the calculation be included as a comment so we don't have to try to figure out 2^1 + 2^2 + 2^3 = 14 seconds?

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NedUMi8--IYZ50WYmvk:b-kmksyg) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

In this case, I think it is fine to move the code to ping the services to the health check classes. Also, could we use [this](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/src/HealthChecks.Aws.S3) instead of writing our own?
</blockquote></details>

OK, I see. I did see that package, yes. Originally, when I wrote the health check, we were not using the Amazon SDK and it seemed awkward to me to mix-and-match in that way. Ultimately, I thought that perhaps regardless it made sense to ping the bucket using what was guaranteed to be the same code that the service is using to increase the usefulness of the health check. Obviously, we don't want a situation where it's reading healthy but isn't because of implementations differences - that could really throw off debugging. Maybe I could write a static method in S3FileStorage so at least the code is all in one file??? Would you rather I just used his health check?
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/machine/85#-NedRu6uD7EspsQsDuDy:-NedV9e39QBEXA6vJirg:b-n80554) ([raw file](https://github.com/sillsdev/machine/blob/c400520d95636cf98a7a61b0737a4435e7944155/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L28)):*

I would put all of this under a single lock.
</blockquote></details>

I wondered about that. OK :thumbsup:.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NedVnznEaxBPrfza7l3:bni6auy) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

OK, I see. I did see that package, yes. Originally, when I wrote the health check, we were not using the Amazon SDK and it seemed awkward to me to mix-and-match in that way. Ultimately, I thought that perhaps regardless it made sense to ping the bucket using what was guaranteed to be the same code that the service is using to increase the usefulness of the health check. Obviously, we don't want a situation where it's reading healthy but isn't because of implementations differences - that could really throw off debugging. Maybe I could write a static method in S3FileStorage so at least the code is all in one file??? Would you rather I just used his health check?
</blockquote></details>

I took a look at the code for the library, and it doesn't have any way to handle the periodic downtime that we want to be able to deal with, so we probably don't want to use it. I think it is fine to not use the shared file service in the health check.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NedZpVnD13WZeHbMo4_:bxcnixn) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

I took a look at the code for the library, and it doesn't have any way to handle the periodic downtime that we want to be able to deal with, so we probably don't want to use it. I think it is fine to not use the shared file service in the health check.
</blockquote></details>

OK, I'll go ahead and copy-paste into the health checks to avoid the retry policy :thumbsup: .

Enkidu93 (commented):
I went ahead and removed the functions I wrote in the `ClearMLService` that were only being used for the health check.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85#-:-NedjuAj9wHw_y7VDPgc:b5c1phv)* status: 2 of 7 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 107 at r2](https://reviewable.io/reviews/sillsdev/machine/85#-Necbn4J5oTjibVU1grW:-NedjmGt5bOuajugY_qs:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/c400520d95636cf98a7a61b0737a4435e7944155/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L107)):*

When will this actually error out?  Can the calculation be included as a comment so we don't have to try to figure out 2^1 + 2^2 + 2^3 = 14 seconds?
</blockquote></details>

Done.
___
*[`src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/machine/85#-NeP9ByeAxOUMTT5wpr4:-NedjnsI9dMkop-UiaKF:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/6ccf9a9c3b559b6a2635a64d86fb68a40538bba0/src/SIL.Machine.AspNetCore/Configuration/IServiceCollectionExtensions.cs#L8)):*

OK, I'll go ahead and copy-paste into the health checks to avoid the retry policy :thumbsup: .
</blockquote></details>

Done.
___
*[`src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/machine/85#-NedRu6uD7EspsQsDuDy:-Nedjs-K7FVvRJnGuOPJ:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/c400520d95636cf98a7a61b0737a4435e7944155/src/SIL.Machine.AspNetCore/Services/S3HealthCheck.cs#L28)):*

I wondered about that. OK :thumbsup:.
</blockquote></details>

Done.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-Nee4SkYCspgLCXj4Lia:bunssa8) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*
> ```Smalltalk
>             using (await _lock.LockAsync())
>                 if (_authToken == "")
>                     _authToken = await GetAuthTokenAsync(cancellationToken);
> ```

Are we sure that we can get multiple valid access tokens for the same user at the same time? I know there are some systems that will invalidate previous access tokens when a new one is created.
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 38 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee2WsO7rzW5gVgIKvs:-Nee2WsO7rzW5gVgIKvt:b-hupopx) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L38)):*
> ```Smalltalk
>     }
> 
>     private async Task<string> GetAuthTokenAsync(CancellationToken cancellationToken)
> ```

Can you move this method to a static ClearML utils class, so it can be used here and in the `ClearMLAuthenticationService` class?
___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 52 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4L1PCBGXyDxzrn5h:-Nee4L1PCBGXyDxzrn5i:bo8u1df) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L52)):*
> ```Smalltalk
>     }
> 
>     private async Task<JsonObject?> CallAsync(
> ```

Can you move this method to a static ClearML utils class, so it can be used here and in the `ClearMLService` class?

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-Nee81eZAPdHQtlFZit4:b8hgipy) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

Are we sure that we can get multiple valid access tokens for the same user at the same time? I know there are some systems that will invalidate previous access tokens when a new one is created.
</blockquote></details>

Running an NMT E2E test while the health checks are running should be sufficient to test this, right?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-Nee7d65DBvOO_0ayOM_:b1lunpw) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

Running an NMT E2E test while the health checks are running should be sufficient to test this, right?
</blockquote></details>

I understand why you didn't use the `ClearMLAuthenticationService` to get the access token, since it has a retry policy on it that can potentially hold up the health check. Go ahead and use it anyway. It will only happen if the server has just started and hasn't retrieved the access token yet.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeBUUyACmI8XtcwUTi:b-l8vb30) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

I understand why you didn't use the `ClearMLAuthenticationService` to get the access token, since it has a retry policy on it that can potentially hold up the health check. Go ahead and use it anyway. It will only happen if the server has just started and hasn't retrieved the access token yet.
</blockquote></details>

So do you still want a static class to hold (just the) `CallAsync` method now?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeFXtbEn5PHUS8gtOk:bby8g64) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

So do you still want a static class to hold (just the) `CallAsync` method now?
</blockquote></details>

We don't need to add it. I retracted those comments.

johnml1135:
It is a little crazy how this simple thing is giving us such pains :-)

johnml1135 (approved):
> Reviewed 5 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeJamA7TltSTjeoVeB:bq79k33) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

We don't need to add it. I retracted those comments.
</blockquote></details>

Oh, I see now. Sorry, I've never encountered Reviewable before here. Apparently, I should take some time to figure it out.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeLU116qPaRnfn2dFF:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

Oh, I see now. Sorry, I've never encountered Reviewable before here. Apparently, I should take some time to figure it out.
</blockquote></details>

Done.

ddaspit (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeRq0V1IugjSkl-VCs:bkkd5ij) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

Done.
</blockquote></details>

We want to get the access token every time the `CallAsync` method is called, otherwise the `ClearMLAuthenticationService` might update the token and we won't get the new token.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/85#-Nee4SkYCspgLCXj4Li_:-NeeTXAq69tcFWCFwY9j:bs7wumb) ([raw file](https://github.com/sillsdev/machine/blob/1a470ea9609b2c32f8fef86a504172cc3df878df/src/SIL.Machine.AspNetCore/Services/ClearMLHealthCheck.cs#L25)):*

We want to get the access token every time the `CallAsync` method is called, otherwise the `ClearMLAuthenticationService` might update the token and we won't get the new token.
</blockquote></details>

Done.

ddaspit (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/85)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Issue #108, opened 2023-10-09
ClearML Refreshes Token too frequently


Enkidu93:
It looks like every time the health check is run, a new instance of `ClearMLAuthenticationService` is created causing Machine to repeatedly refresh the auth token. I don't think this is how it ought to work - it should be a proper singleton.

johnml1135:
I agree - it should (ideally) be a singleton that refreshes every day (or every hour?) and when it does so, it has a few attempts if there is a failure - and allows other requests to use the old authentication until new authentication is established.

PR #114, opened 2023-10-18
Fixes ClearML Refreshes Token too frequently #108

Enkidu93:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/114)


Enkidu93:
> I don't think this will apply the error policy to the HTTP client that is injected into the services. We can't use typed HTTP clients. We need to use a different approach to inject the HTTP client. I have an idea for how I would like to refactor all of this to work properly. I am going to submit a separate PR with the fix. You can close this PR.

That's possible, I suppose. I'm confused though: We were using typed HTTP clients before (I hadn't changed that - just the order), and it was working. Or do you just mean in the context of having singleton services? I thought about moving to the HttpClientFactory paradigm which is a lot of folks recommend it seems, but OK.

ddaspit:
Yes, they simply won't work for singleton services.

PR #143, opened 2023-11-28
Key terms support

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/225

Parallel PR: https://github.com/sillsdev/serval/pull/226

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)


johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMgg_y4wSeK5Fgavuv:-NkMgg_y4wSeK5Fgavuw:b-ovaauh) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L23)):*
> ```Smalltalk
>                 if (keyTermsFileEntry is null)
>                 {
>                     throw new ArgumentException(
> ```

Should we really crash if there is no keyterms file?  That appears excessive.

Enkidu93:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMgg_y4wSeK5Fgavuv:-NkMh3Nx8FQIPEokJOi1:b-mwblms) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L23)):*

Should we really crash if there is no keyterms file?  That appears excessive.
</blockquote></details>

It's caught upstream. I think it's appropriate that if you try to make a Paratext*KeyTerms*Corpus with no key terms file, it throws an exception.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMgg_y4wSeK5Fgavuv:-NkMhizx7S4l82cTsvxq:b-iiadh1) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L23)):*

It's caught upstream. I think it's appropriate that if you try to make a Paratext*KeyTerms*Corpus with no key terms file, it throws an exception.
</blockquote></details>

ok - as long as it just keeps going.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMho3LBqa2pMlhK6qq:-NkMho3LBqa2pMlhK6qr:b-5863iy) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L61)):*
> ```Smalltalk
>                     Regex rx = new Regex(@"\[(.+?)\]", RegexOptions.Compiled);
>                     Match match = rx.Match(gloss);
>                     if (match.Success)
> ```

There should be a nice set of 20 or so tests to check this.  I would recommend breaking out the parsing section as a separate function that takes a list of keyTermsElements and returns rows.  That should be more testable than having to create a bunch of xml files.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMho3LBqa2pMlhK6qq:-NkMiGPO6PHAIODSbqrK:bvo150a) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L61)):*

There should be a nice set of 20 or so tests to check this.  I would recommend breaking out the parsing section as a separate function that takes a list of keyTermsElements and returns rows.  That should be more testable than having to create a bunch of xml files.
</blockquote></details>

The 20 or so tests would be the official documentation of the logic.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMho3LBqa2pMlhK6qq:-NkMiOfI0NjmZsO_I3OF:b-5kcyoj) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L61)):*

The 20 or so tests would be the official documentation of the logic.
</blockquote></details>

Or, you could have a function that just takes a single element and returns a list of glosses - that may be better.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiZRABJjqzLER9xjH:-NkMiZRABJjqzLER9xjI:b-on33pd) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L81)):*
> ```Smalltalk
>         }
> 
>         private static string StripParens(string termString, char left = '(', char right = ')')
> ```

This needs a comment explaining what it does and a few tests to confirm behavior.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMho3LBqa2pMlhK6qq:-NkMiiCdAGRc5zZ9rTrh:bfx9fi) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L61)):*

Or, you could have a function that just takes a single element and returns a list of glosses - that may be better.
</blockquote></details>

And the function can be static.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 133 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiutyAtXW7vZBjhnP:-NkMiutyAtXW7vZBjhnQ:b-ykdfrd) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L133)):*
> ```Smalltalk
>                 if (
>                     buildOptionsObject is not null
>                     && buildOptionsObject["use_key_terms"] is not null
> ```

Using keyterms should be default.  Possibly you could switch the logic?  What would be the best way to do this @ddaspit?

johnml1135:
This needs more testing.

johnml1135 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMhF8n-7iuZDtEWGAw:-NkMhF8n-7iuZDtEWGAx:b-2c9v0l) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*
> ```Smalltalk
>                 foreach (XElement element in keyTermsElements)
>                 {
>                     if ((element.Element("Category")?.Value ?? "") != "PN")
> ```

This logic should be documented in a wiki.

ddaspit (changes requested):
We want this functionality for all engine types, including SMT, so the code to create a `ParatextKeyTermsCorpus` should be moved to `CorpusService.CreateTextCorpus`. Once this functionality is no longer specific to NMT, it is no longer appropriate to use build options to configure this. Build options are engine-specific options. A new property should be added to `StartBuildRequest` for the `StartBuild` gRPC endpoint to toggle this functionality.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143#-:-NkRdnyNCRpT1pdJi4th:b37h5p8)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93)

Enkidu93 (commented):
Logic has been moved.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143#-:-Nk_vr2N9GVpRrJEp_hi:b-sghlha)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMhF8n-7iuZDtEWGAw:-NkQKSKg3VvYSSVjQ-xy:bitn646) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*

This logic should be documented in a wiki.
</blockquote></details>

This logic has been removed since we're using TermRenderings.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 61 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMho3LBqa2pMlhK6qq:-NkQKNf-0X1hk7pti-VP:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L61)):*

And the function can be static.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 81 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiZRABJjqzLER9xjH:-NkQKQfs9-9htfv8wLZr:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L81)):*

This needs a comment explaining what it does and a few tests to confirm behavior.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 133 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiutyAtXW7vZBjhnP:-NkQK9Nq8Bomyt5oVO4a:b-i7x1f0) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L133)):*

Using keyterms should be default.  Possibly you could switch the logic?  What would be the best way to do this @ddaspit?
</blockquote></details>

OK, it's easy enough to make that the default; I just didn't know that it was supposed to be a default feature. I've left it as-is. Please confirm that we'd want this to be the default behavior and I can make that change.

Enkidu93:
The added testing also fixes #144 .

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-Nka8fwy9nF-ahpOtaSJ:-Nka8fwzEJoUyZE_8XgD:b4hblwd) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L23)):*
> ```Smalltalk
>     }
> 
>     public ParatextKeyTermsCorpus? CreateKeyTermsCorpus(IReadOnlyList<CorpusFile> files)
> ```

This appears to not have code coverage: https://app.codecov.io/gh/sillsdev/machine/pull/143?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=sillsdev

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 133 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiutyAtXW7vZBjhnP:-Nka9ATr8LNiC92Ew42y:btqnjwx) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L133)):*

OK, it's easy enough to make that the default; I just didn't know that it was supposed to be a default feature. I've left it as-is. Please confirm that we'd want this to be the default behavior and I can make that change.
</blockquote></details>

@ddaspit - do you have any objection to making it default?

johnml1135 (commented):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @Enkidu93)

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-Nka8fwy9nF-ahpOtaSJ:-NkaFvgvEocR1m7nGlYT:b-2k5flm) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L23)):*

This appears to not have code coverage: https://app.codecov.io/gh/sillsdev/machine/pull/143?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=sillsdev
</blockquote></details>

Thank you! Previously, this file had 0% coverage. The generic `CreateTextCorpus` gets tested in the `NmtPreprocessBuildJobTests`. It would be a bit forced and odd to test this there as well. I could also make it its own `CorpusServiceTests` but there's just so little to test. This **is** tested in the Serval Paratext E2E test. If you'd really like me to, I can incorporate a test now for this. It may be worthwhile to open up a separate issue that involves systematically going through the existing code coverage reports and trying to improve our coverage.

ddaspit (changes requested):
I apologize, but I am asking you to change the design again. See my comments below.

> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143#-:-NkaYLDo5wvU93yZ1_9Z:b-ldxjm)* status: all files reviewed, 8 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 15 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSKONAjeTgkTAmtuL:-NkaSKONAjeTgkTAmtuM:b-zb67k9) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L15)):*
> ```Smalltalk
>         public string BiblicalTermsType { get; set; }
> 
>         public ParatextKeyTermsCorpus(string projectDir)
> ```

I would rename this to `fileName`.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaT-SjCPgP2hAMdRAR:-NkaT-SjCPgP2hAMdRAS:b-nszsfv) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L23)):*
> ```Smalltalk
>                 if (termsFileEntry is null)
>                 {
>                     throw new ArgumentException(
> ```

Instead of throwing an exception, we should just return an empty corpus.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSZ_55Yklg_oUO5zN:-NkaSZ_69TJONbT1A2LY:bayrw6m) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*
> ```Smalltalk
>                 foreach (XElement element in termsElements)
>                 {
>                     string id = element.Attribute("Id").Value;
> ```

We need some way to filter by category. This means that we will need to pull in the Biblical terms list.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 56 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaWGgJ618iMVkpznzy:-NkaWGgJ618iMVkpznzz:b-i7m1cb) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L56)):*
> ```Smalltalk
>                     string id = element.Attribute("Id").Value;
>                     id = id.Replace("\n", "&#xA");
>                     string gloss = element.Element("Renderings").Value;
> ```

I would rename this to `rendering` since it could be confused with the term gloss.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 60 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaUiI36nPingQUHer5:-NkaUiI36nPingQUHer6:bhtv4kg) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L60)):*
> ```Smalltalk
>                     rows.Add(new TextRow("KeyTerms", id) { Segment = glosses });
>                 }
>                 IText text = new MemoryText("KeyTerms", rows);
> ```

If the text id is the Biblical terms list, then we won't need to add an extra check to see if the source and target are using the same list.

If we do this, then we don't need to perform the special check to see if the lists match, so we can treat this just like any other corpus. I think we should change `CorpusService.CreateTextCorpus` to return a dictionary of corpora, which was one of the options we discussed previously. The key would be an enum that defines the type of data in the corpus, i.e. `Text` and `Terms`. You can match up the source and target corpora using the enum values. You can also use the enum value to remove the key terms if they are disabled.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 88 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaTMYl9t6U0O_iUoZ-:-NkaTMYl9t6U0O_iUoZ0:bsbdtj4) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L88)):*
> ```Smalltalk
>                     buildOptionsObject is not null
>                     && buildOptionsObject["use_key_terms"] is not null
>                     && buildOptionsObject["use_key_terms"]!.ToString() == "true"
> ```

I think this can be cast to a `bool`.

Enkidu93:
OK :face_with_spiral_eyes:

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-Nka8fwy9nF-ahpOtaSJ:-NkbI4pm9YuTTnJSE2no:b-vk5mex) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L23)):*

Thank you! Previously, this file had 0% coverage. The generic `CreateTextCorpus` gets tested in the `NmtPreprocessBuildJobTests`. It would be a bit forced and odd to test this there as well. I could also make it its own `CorpusServiceTests` but there's just so little to test. This **is** tested in the Serval Paratext E2E test. If you'd really like me to, I can incorporate a test now for this. It may be worthwhile to open up a separate issue that involves systematically going through the existing code coverage reports and trying to improve our coverage.
</blockquote></details>

If it is tested in E2E testing (and no logic of note), then I am ok ignoring it.

johnml1135 (commented):
> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit and @Enkidu93)

Enkidu93:
> _[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 23 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-Nka8fwy9nF-ahpOtaSJ:-NkbI4pm9YuTTnJSE2no:b-vk5mex) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L23)):_
> Previously, Enkidu93 (Eli C. Lowry) wrote…
> 
> If it is tested in E2E testing (and no logic of note), then I am ok ignoring it.

OK, cool. And it is slightly more covered now anyways after the re-redesign.

ddaspit (commented):
> Reviewed 6 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 15 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSKONAjeTgkTAmtuL:-NkbN6A0-GKPBCkEw2k0:b-mnanqa) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L15)):*

Done. Would you like me to rename `"projectDir"` -> `"fileName"` in `ParatextTextCorpus` as well then? It was already inconsistent language, but "projectDir" seemed the more appropriate of the two.
</blockquote></details>

I think you are referring to `ParatextBackupTextCorpus`. That class already uses `fileName`.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSZ_55Yklg_oUO5zN:-NkbOHQF2XhX0Mlo_PCT:bv08fzv) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*

Done.
</blockquote></details>

We still need to deal with the case where the project is using one of the predefined Biblical terms lists. I think that means we need to include the predefined XML files as embedded resources in the `SIL.Machine` assembly. You can see an example of this with the [embedded `usfm.sty` file](https://github.com/sillsdev/machine/blob/5a488c01452f8cfb30214cdb6d90ab8a517ed623/src/SIL.Machine/Corpora/UsfmStylesheet.cs#L113). We can talk next week, if you want more details.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 74 at r3](https://reviewable.io/reviews/sillsdev/machine/143#-NkbNd-rAOMHK4RudSf_:-NkbNd-rAOMHK4RudSfa:bksn40y) ([raw file](https://github.com/sillsdev/machine/blob/32484a105eb08893dc78a75e90afafb140e73db9/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L74)):*
> ```Smalltalk
>                     id = id.Replace("\n", "&#xA");
>                     string rendering = element.Element("Renderings").Value;
>                     IReadOnlyList<string> glosses = GetRenderings(rendering);
> ```

You missed this variable name.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 133 at r1](https://reviewable.io/reviews/sillsdev/machine/143#-NkMiutyAtXW7vZBjhnP:-NkbQ9IHDFV2DMdJ-TDF:ber09kt) ([raw file](https://github.com/sillsdev/machine/blob/aade573dce7bd68cc5268cdc59001d0af42c8d9c/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L133)):*

@ddaspit - do you have any objection to making it default?
</blockquote></details>

Yes, this should be turned on by default.

johnml1135 (commented):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Enkidu93)

johnml1135 (commented):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Enkidu93)

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 94 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkp6ZkWCdvyD-a8q1U6:-Nkp7qWLDSe3vdcvpKmf:bk1ktbr) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L94)):*

Do we test that the keyterms are included and not included?  Where is the test?
</blockquote></details>

I can add a second `NmtBuildJobTests` test if you like that's Paratext-specific. That would be the place to add it. Would that do?

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 94 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkp6ZkWCdvyD-a8q1U6:-NkpDT4LCTmhY_YFPVlc:bnd4972) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L94)):*

I can add a second `NmtBuildJobTests` test if you like that's Paratext-specific. That would be the place to add it. Would that do?
</blockquote></details>

Ideally something in this repo that is a unit test, merely testing that the tokens are not included with the right buildOptions.

ddaspit (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 15 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSKONAjeTgkTAmtuL:-Nkq-QNP0AlN3M1ETQID:b-i25cfw) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L15)):*

No, I'm referring to [this class](https://github.com/sillsdev/machine/blob/master/src/SIL.Machine/Corpora/ParatextTextCorpus.cs). Maybe it's unused? In any case, it seems to me all three should be consistent.
</blockquote></details>

`ParatextTextCorpus` is used to load a Paratext project directory. That is why it uses the parameter `projectDir`. `ParatextBackupTextCorpus` is used to load a zipped Paratext project backup. That actually raises an issue with the name of this class. It should be called `ParatextBackupTermsCorpus`, since it is used to load key terms from a zipped Paratext project.
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSZ_55Yklg_oUO5zN:-Nkq0Rui-V0plp1zLrI0:bqmgg8l) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*

I see. So should I include the Biblical terms lists that are listed in Paratext by default? I'd have to go back to my other computer to see what all their names are (Paratext is not happy on my Linux box), but something like "MajorBiblicalTerms", "AllBiblicalTerms", etc. Or which ones should I embed? And than that would require adding analogous lines [here](https://github.com/sillsdev/machine/blob/5a488c01452f8cfb30214cdb6d90ab8a517ed623/src/SIL.Machine/SIL.Machine.csproj#L10C1-L18C14), yes? Anything else?
</blockquote></details>

There are four terms lists that you should embed: `AllBliblicalTerms`, `BiblicalTerms`, `BiblicalTermsP6NT`, and `BiblicalTermsSILNT`. You can find them in the S3 bucket at `Paratext/terms`.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 91 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq1bEJ0gKMLtSkxg6Y:-Nkq1bEK5MrjKwGbi6lW:b-mxbl4f) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L91)):*
> ```Smalltalk
>                 if (
>                     (
>                         buildOptionsObject is null
> ```

It would be better to use null coalescing and null conditional operators here, i.e. `(bool?)buildOptionsObject?["use_key_terms"] ?? true`.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 15 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSKONAjeTgkTAmtuL:-Nkq6HjS0jhYzHu2kqPd:b-pmjo3y) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L15)):*

`ParatextTextCorpus` is used to load a Paratext project directory. That is why it uses the parameter `projectDir`. `ParatextBackupTextCorpus` is used to load a zipped Paratext project backup. That actually raises an issue with the name of this class. It should be called `ParatextBackupTermsCorpus`, since it is used to load key terms from a zipped Paratext project.
</blockquote></details>

I see the rationale now: It is properly a file even though it's a zipped folder. That makes sense. I'll go ahead and update the name. 
___
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSZ_55Yklg_oUO5zN:-Nkq6VhO6P8zCGoFXOdV:b-ab8oj3) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*

There are four terms lists that you should embed: `AllBliblicalTerms`, `BiblicalTerms`, `BiblicalTermsP6NT`, and `BiblicalTermsSILNT`. You can find them in the S3 bucket at `Paratext/terms`.
</blockquote></details>

OK, I will embed those. Just to make sure I'm understanding, if a project is using one of these, the name of the file will appear in the Settings.xml as the BiblicalTermsListSetting value?
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 91 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq1bEJ0gKMLtSkxg6Y:-Nkq78nY8nO2MRuh4O0s:b-oz8tym) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L91)):*

It would be better to use null coalescing and null conditional operators here, i.e. `(bool?)buildOptionsObject?["use_key_terms"] ?? true`.
</blockquote></details>

Good call.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 94 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkp6ZkWCdvyD-a8q1U6:-Nkq7BoVBEuy1Wswp5_j:b-osr5zy) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L94)):*

Ideally something in this repo that is a unit test, merely testing that the tokens are not included with the right buildOptions.
</blockquote></details>

I'll go ahead and make a Paratext test as well in the `NmtBuildJobTests.cs`

Enkidu93:
*[`src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs` line 54 at r2](https://reviewable.io/reviews/sillsdev/machine/143#-NkaSZ_55Yklg_oUO5zN:-NkqDq7lCt3Pay0s4op0:bnyeg7l) ([raw file](https://github.com/sillsdev/machine/blob/67e2a052690b2f5160699ce22cb80df491f1bf49/src/SIL.Machine/Corpora/ParatextKeyTermsCorpus.cs#L54)):*

OK, I will embed those. Just to make sure I'm understanding, if a project is using one of these, the name of the file will appear in the Settings.xml as the BiblicalTermsListSetting value?
</blockquote></details>

Nevermind, I figured it out.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 94 at r4](https://reviewable.io/reviews/sillsdev/machine/143#-Nkp6ZkWCdvyD-a8q1U6:-NkqWlcm6vvJdbX72-XG:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/5e777b68e1915410a8dac1d845c92842ea9ebb59/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L94)):*

I'll go ahead and make a Paratext test as well in the `NmtBuildJobTests.cs`
</blockquote></details>

Done.

johnml1135 (approved):
> Reviewed 1 of 2 files at r5, 9 of 10 files at r6, 11 of 11 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> Reviewed 2 of 2 files at r5, 9 of 10 files at r6, 11 of 11 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 93 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_-BjA-mC-CMmiuxm:-Nkq_-BjA-mC-CMmiuxn:b-vw8brm) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L93)):*
> ```Smalltalk
>                     ?? true
>                         && sourceCorpora.ContainsKey(CorpusType.Term)
>                         && sourceCorpora[CorpusType.Term] is not null
> ```

There is no need to check if this is not `null`.
___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferBuildJob.cs` line 71 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_G6yEkF5h2ihvHjq:-Nkq_G6yEkF5h2ihvHjr:by6sl1m) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/SmtTransferBuildJob.cs#L71)):*
> ```Smalltalk
>             if (
>                 (bool?)buildOptionsObject?["use_key_terms"]
>                 ?? true && scs[CorpusType.Term] is not null && tcs[CorpusType.Term] is not null
> ```

You should use `ContainsKey` like you did in the NMT build.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-NkqWukO0nFqyVWvzK27:bjwo8q0) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*
> ```Smalltalk
> 
>                 XDocument biblicalTermsDoc;
>                 if (biblicalTermsFileEntry != null)
> ```

You should use `BiblicalTermsListSetting` to determine if the project is using a custom terms list. The format of the setting is `<ListType>:<ProjectId>:<TermsFileName>`. For a project terms list, the `<ListType>` will be `Project` or `OtherProject`. If the type is `Project`, then the terms list file will be named `ProjectBiblicalTerms.xml`. If it is `OtherProject`, then it will named `<TermsFileName>` from the setting. The `<ListType>` for the predefined lists are `Major`, `All`, `SilNt`, and `Pt6`. The `<ProjectId>` will be blank for predefined lists. There are other list types that we cannot use, so we should ignore any other list type.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-NkqZeIo-yHI6j8F7bYD:ba9uhuk) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*
> ```Smalltalk
>                                 .Element("Category")
>                                 ?.Value ?? ""
>                         ) != "PN"
> ```

We should pass in the category that we want to filter on to the constructor. If it is `null`, then we return all terms.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 17 at r8](https://reviewable.io/reviews/sillsdev/machine/143#-NkuBMqa3PdBWs7CCNJX:-NkuBMqa3PdBWs7CCNJY:banhfpo) ([raw file](https://github.com/sillsdev/machine/blob/64a7b05d68df80a22f33405194c035e036f44e0c/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L17)):*
> ```Smalltalk
>         {
>             corpora[CorpusType.Text] = new ParatextBackupTextCorpus(files[0].Location);
>             corpora[CorpusType.Term] = new ParatextBackupTermsCorpus(files[0].Location, new string[] { "PN" });
> ```

I am assuming that this keeps using proper nouns only for the web service, but not the core library.  That appears to be a good decision.

johnml1135 (commented):
> Reviewed 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93)

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 93 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_-BjA-mC-CMmiuxm:-Nkr2QiS2EsXzEGeh25j:bvty9s6) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L93)):*

There is no need to check if this is not `null`.
</blockquote></details>

But since `CreateTextCorpus` might return `null` for the `ParatextBackupTermsCorpus`, couldn't this value be null at which point `AlignRows` would fail?
___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferBuildJob.cs` line 71 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_G6yEkF5h2ihvHjq:-Nkr2PMW3RhD_jdDKnzk:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/SmtTransferBuildJob.cs#L71)):*

You should use `ContainsKey` like you did in the NMT build.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-Nkr3K4H0DAscWpBt9TN:b-7hkw2t) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*

You should use `BiblicalTermsListSetting` to determine if the project is using a custom terms list. The format of the setting is `<ListType>:<ProjectId>:<TermsFileName>`. For a project terms list, the `<ListType>` will be `Project` or `OtherProject`. If the type is `Project`, then the terms list file will be named `ProjectBiblicalTerms.xml`. If it is `OtherProject`, then it will named `<TermsFileName>` from the setting. The `<ListType>` for the predefined lists are `Major`, `All`, `SilNt`, and `Pt6`. The `<ProjectId>` will be blank for predefined lists. There are other list types that we cannot use, so we should ignore any other list type.
</blockquote></details>

OK, see what I've done and tell me how you'd like it to be different. I think I've incorporated what you've said appropriately. I am little confused though because when I added Biblical terms to the Paratext project I've been playing with to get test data, it kept the type "Major" but still had a "ProjectBiblicalTerms.xml" file with just the terms I had starred, so shouldn't we honor that project-specific file if present?
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-Nkr2qlv-Mnlg1R1kyks:b33vrzj) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*

We should pass in the category that we want to filter on to the constructor. If it is `null`, then we return all terms.
</blockquote></details>

I've made it a list of category strings (if you want just a singular one, I'm happy to change it, but I figured that might be useful). Also, since we're passing it to the constructor, would you like it to be passed into `CreateTextCorpus` as well (that way, we could configure it externally - e.g. from Serval) or just have a magic value in `CreateTextCorpus`. The former seems more useful, but it does seem a little odd to pass term categories to `CreateTextCorpus`.

ddaspit (commented):
> Reviewed 5 of 5 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 93 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_-BjA-mC-CMmiuxm:-Nkuz3yy3kPzEU26g_zL:b-jcsqqv) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L93)):*

But since `CreateTextCorpus` might return `null` for the `ParatextBackupTermsCorpus`, couldn't this value be null at which point `AlignRows` would fail?
</blockquote></details>

I don't think there is any way that `sourceCorpora[CorpusType.Term]` or `targetCorpora[CorpusType.Term]` could be `null`. There isn't anywhere where a value in the dictionary is set to `null`.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-Nkv4W7P852zDbkQmleY:b-5vdvk1) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*

OK, see what I've done and tell me how you'd like it to be different. I think I've incorporated what you've said appropriately. I am little confused though because when I added Biblical terms to the Paratext project I've been playing with to get test data, it kept the type "Major" but still had a "ProjectBiblicalTerms.xml" file with just the terms I had starred, so shouldn't we honor that project-specific file if present?
</blockquote></details>

The Biblical terms feature in PT is a bit confusing. When you "star" a term, it will be added to the project terms list. PT will create a `ProjectBiblicalTerms.xml` file in the project folder, but that doesn't mean that the project is configured to use it. You have to specifically associate the project terms list with the project to make it the active list. You can do this in the project settings under the "Association" tab. 
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-NkuyPpk0z8QOg9mQSue:b-5z8tpj) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*

I've made it a list of category strings (if you want just a singular one, I'm happy to change it, but I figured that might be useful). Also, since we're passing it to the constructor, would you like it to be passed into `CreateTextCorpus` as well (that way, we could configure it externally - e.g. from Serval) or just have a magic value in `CreateTextCorpus`. The former seems more useful, but it does seem a little odd to pass term categories to `CreateTextCorpus`.
</blockquote></details>

A list is good.

johnml1135 (commented):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/CorpusService.cs` line 17 at r8](https://reviewable.io/reviews/sillsdev/machine/143#-NkuBMqa3PdBWs7CCNJX:-NkvAViL5uC9bhyMtyqG:bp8eu3) ([raw file](https://github.com/sillsdev/machine/blob/64a7b05d68df80a22f33405194c035e036f44e0c/src/SIL.Machine.AspNetCore/Services/CorpusService.cs#L17)):*

I am assuming that this keeps using proper nouns only for the web service, but not the core library.  That appears to be a good decision.
</blockquote></details>

Yes, that's correct.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-Nkv9ogQ1l--pM-pI1E4:b-yfic3g) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*

The Biblical terms feature in PT is a bit confusing. When you "star" a term, it will be added to the project terms list. PT will create a `ProjectBiblicalTerms.xml` file in the project folder, but that doesn't mean that the project is configured to use it. You have to specifically associate the project terms list with the project to make it the active list. You can do this in the project settings under the "Association" tab.
</blockquote></details>

OK, I see. Seems like Paratext is a bit bloated with so many features and menus. So is this logic what you were looking for then?
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-NkvA3gb8zKyHjAV46a9:brslq7c) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*

A list is good.
</blockquote></details>

I'll assume then that where "PN" is injected it appropriate.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 93 at r6](https://reviewable.io/reviews/sillsdev/machine/143#-Nkq_-BjA-mC-CMmiuxm:-NkvCs8q55AGFG4Dqako:b2xasyz) ([raw file](https://github.com/sillsdev/machine/blob/9e413e55a74566405b2df770820b537e216ead77/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L93)):*

I don't think there is any way that `sourceCorpora[CorpusType.Term]` or `targetCorpora[CorpusType.Term]` could be `null`. There isn't anywhere where a value in the dictionary is set to `null`.
</blockquote></details>

Yes, I was forgetting that it was just returning an empty corpus not returning null on failure (since it's the key terms constructor itself). Done.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: 24 of 25 files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-NkvDOPU1omA3CIBnlV1:b-al03mt) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*

OK, I see. Seems like Paratext is a bit bloated with so many features and menus. So is this logic what you were looking for then?
</blockquote></details>

For a project terms list, you just need to check if the list type is `Project` and the project id matches the `Name` field in the project settings.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-NkvCALHB_Sype6bRAiP:bxi7t4j) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*

I'll assume then that where "PN" is injected it appropriate.
</blockquote></details>

Yes, that looks correct.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 80 at r10](https://reviewable.io/reviews/sillsdev/machine/143#-NkvEJTw1O4kao6uyxK7:-NkvEJTw1O4kao6uyxK8:bevlvl1) ([raw file](https://github.com/sillsdev/machine/blob/d0509f3b938414cda0192d4ed70596f43087e1ca/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L80)):*
> ```Smalltalk
>                             termCategories.Count() > 0
>                             && !termCategories.Contains(
>                                 biblicalTermsDoc
> ```

This is potentially expensive, since it is searching through the whole terms list XML for every term rendering. It would probably be better to create a dictionary from the terms list that maps the term id to the category.

ddaspit (commented):
> Reviewed 1 of 1 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 80 at r10](https://reviewable.io/reviews/sillsdev/machine/143#-NkvEJTw1O4kao6uyxK7:-NkvYSks4w3vXy-B9JJC:b-o2jgb4) ([raw file](https://github.com/sillsdev/machine/blob/d0509f3b938414cda0192d4ed70596f43087e1ca/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L80)):*

This is potentially expensive, since it is searching through the whole terms list XML for every term rendering. It would probably be better to create a dictionary from the terms list that maps the term id to the category.
</blockquote></details>

You should use `TryGetValue`, so you only perform one dictionary lookup.

Enkidu93:
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 80 at r10](https://reviewable.io/reviews/sillsdev/machine/143#-NkvEJTw1O4kao6uyxK7:-NkvgR087Bj98OKbK5gX:bhj4jhi) ([raw file](https://github.com/sillsdev/machine/blob/d0509f3b938414cda0192d4ed70596f43087e1ca/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L80)):*

You should use `TryGetValue`, so you only perform one dictionary lookup.
</blockquote></details>

Still not passing a test - checking the logic again now.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: 23 of 26 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 49 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqWukO0nFqyVWvzK26:-NkvgIJV0OKNvHnAE0_u:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L49)):*

For a project terms list, you just need to check if the list type is `Project` and the project id matches the `Name` field in the project settings.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/machine/143#-NkqZeIo-yHI6j8F7bYC:-NkvkTyU6Ip8XPBd_gia:b-89q77k) ([raw file](https://github.com/sillsdev/machine/blob/90980672507c43c2df42625513de7cc204692f68/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L84)):*

Yes, that looks correct.
</blockquote></details>

Cool.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 80 at r10](https://reviewable.io/reviews/sillsdev/machine/143#-NkvEJTw1O4kao6uyxK7:-NkvkSTU-_2hZfdSGmmT:b-i64o6i) ([raw file](https://github.com/sillsdev/machine/blob/d0509f3b938414cda0192d4ed70596f43087e1ca/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L80)):*

Still not passing a test - checking the logic again now.
</blockquote></details>

Fixed the test data.

johnml1135 (commented):
> Reviewed 3 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

ddaspit (commented):
> Reviewed 3 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 61 at r12](https://reviewable.io/reviews/sillsdev/machine/143#-Nkw7X_R1iOinbATd7_v:-Nkw7X_R1iOinbATd7_w:bhbv1b2) ([raw file](https://github.com/sillsdev/machine/blob/7be4a6b27d8238f85ce08a2a7651052bb1e1dde0/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L61)):*
> ```Smalltalk
>                     {
>                         biblicalTermsDoc = XDocument.Load(keyTermsFile);
>                         termIdToCategoryDictionary = biblicalTermsDoc
> ```

Can you move this code to a static method so it isn't repeated?
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 69 at r12](https://reviewable.io/reviews/sillsdev/machine/143#-Nkw7ewM0IWltVowH0A9:-Nkw7ewM0IWltVowH0AA:bpli352) ([raw file](https://github.com/sillsdev/machine/blob/7be4a6b27d8238f85ce08a2a7651052bb1e1dde0/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L69)):*
> ```Smalltalk
>                 else
>                 {
>                     if (
> ```

This can be an `else if` statement.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 61 at r12](https://reviewable.io/reviews/sillsdev/machine/143#-Nkw7X_R1iOinbATd7_v:-NkzUOo6-URCg9bFtIBK:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/7be4a6b27d8238f85ce08a2a7651052bb1e1dde0/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L61)):*

Can you move this code to a static method so it isn't repeated?
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 69 at r12](https://reviewable.io/reviews/sillsdev/machine/143#-Nkw7ewM0IWltVowH0A9:-NkzUPFp1Ob12uQYJATP:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/7be4a6b27d8238f85ce08a2a7651052bb1e1dde0/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L69)):*

This can be an `else if` statement.
</blockquote></details>

Done.

johnml1135 (approved):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/143)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

Enkidu93:
And thank you to you for being persistent in your reviewing.

PR #161, opened 2024-01-26
Added chapter-level filtering

Enkidu93:
Addresses  https://github.com/sillsdev/serval/issues/150

I'm noticing that (as far as I can tell) the "trainOn" logic is not present in SMT jobs. It makes sense why pretranslate logic wouldn't be there, but should we incorporate the trainOn logic? I won't add the biblicalRange logic to SMT for training until it's been verified that that was a mistake.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)


Enkidu93:
My csharpier was broken (+ I forgot about the mismatched .NET versions). I'll go ahead and take care of that.

ddaspit (changes requested):
Rather than require that all engine implementations be able to parse the Scripture range string, I think it would be better to perform the parsing in Serval. Serval will already have a dependency on SIL.Machine for USFM generation, so we can do the same for the Scripture range parsing.

> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np5xlGcA9SNcYUleOPs:b-12uptx)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/BiblicalRangeStringParser.cs` line 0 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np5yUS-2WNhBLadpBYk:-Np5yUS-2WNhBLadpBYl:biyfnly) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/BiblicalRangeStringParser.cs#L0)):*
I order to better match up with Machine.py:

- this class should be moved to `SIL.Machine/Scripture`
- Turned into a static class and renamed to `ScriptureRangeParser`
- The `Parse` should be made static and renamed to `GetChapters`.

Enkidu93 (commented):
And then pass a dictionary rather than a string from Serval to Machine? You're not suggesting the other filtering logic is somehow put into Serval, right?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np60pZP1oQF-N7WlJHG:b-70aapc)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/BiblicalRangeStringParser.cs` line  at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np5yUS-2WNhBLadpBYk:-Np60nTR7OEnSO5nbGyN:by6kb2t) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/BiblicalRangeStringParser.cs#L)):*

I order to better match up with Machine.py:

- this class should be moved to `SIL.Machine/Scripture`
- Turned into a static class and renamed to `ScriptureRangeParser`
- The `Parse` should be made static and renamed to `GetChapters`.
</blockquote></details>

That seems reasonable.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 141 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np64GLY3sz9Ep_mY3aG:-Np64GLZ1BU56I9bzMe7:bi93x56) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L141)):*
> ```Smalltalk
>                         var parser = new BiblicalRangeStringParser(stc.Versification);
>                         if(corpus.TrainOnBiblicalRange != null && corpus.TrainOnBiblicalRange != ""){
>                             Dictionary<string, List<int>> trainOnBiblicalRangeChapters = parser.Parse(corpus.TrainOnBiblicalRange); //TODO calculate once
> ```

So, it's calculated for every row of text?  That should be fixed...

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 142 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np64ZI-CXnYeT_OAl1H:-Np64ZI-CXnYeT_OAl1I:b-o5b0pj) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L142)):*
> ```Smalltalk
>                         if(corpus.TrainOnBiblicalRange != null && corpus.TrainOnBiblicalRange != ""){
>                             Dictionary<string, List<int>> trainOnBiblicalRangeChapters = parser.Parse(corpus.TrainOnBiblicalRange); //TODO calculate once
>                             isInTrainOnRange = rowChaptersPerBook.Join(trainOnBiblicalRangeChapters, rcpb => rcpb.Key, tobrc => tobrc.Key, (rcbp, tobrc) =>
> ```

So, to be clear, if the verse is in either the source or the target versification, it will work?  Is that what we want?

ddaspit (commented):
Yes, we would just parse the string into a data structure that we pass to Machine.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np65suR7sDWcW4m4bB8:b-annzwg)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 147 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np66eeb0lx3PGAdSE_s:-Np66eeb0lx3PGAdSE_t:br61g0x) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L147)):*
> ```Smalltalk
>                         }
>                         if(corpus.PretranslateBiblicalRange != null && corpus.PretranslateBiblicalRange != ""){
>                             Dictionary<string, List<int>> pretranslateBiblicalRangeChapters = parser.Parse(corpus.PretranslateBiblicalRange);
> ```

Also this one needs pulled out of the loop too.

Enkidu93:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 147 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np66eeb0lx3PGAdSE_s:-Np67VX02QfYkJS76Z65:b8vr72j) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L147)):*

Also this one needs pulled out of the loop too.
</blockquote></details>

Yep! Thank you

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 137 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np674nd5kyKjZkcg6ZV:-Np674ne3ORRkbyetnZg:bmvsgyl) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L137)):*
> ```Smalltalk
>                     bool isInTrainOnRange = false;
>                     bool isInPretranslateRange = false;
>                     if(targetCorpora[CorpusType.Text] is ScriptureTextCorpus stc && row.Refs.All(r => r is VerseRef)){
> ```

We should probably only do these checks if there is a biblical range - if not, just skip it.  Less computation?

johnml1135:
*[`tests/SIL.Machine.AspNetCore.Tests/Services/data/paratext2/TermRenderings.xml` line 2 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np6HgwcD5wHTcwGIxV2:-Np6HgwcD5wHTcwGIxV3:b-o91jz8) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/tests/SIL.Machine.AspNetCore.Tests/Services/data/paratext2/TermRenderings.xml#L2)):*
> ```XML
> <TermRenderingsList>
>   <TermRendering Id="Abba" Guess="false">
> ```

What is the purpose of these term renderings and other files?

johnml1135 (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit and @Enkidu93)

___
*[`tests/SIL.Machine.AspNetCore.Tests/Services/BiblicalRangeStringParserTests.cs` line 73 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np6H4zi0IbM8UKzxHbZ:-Np6H4zj25VfDlLNQf9d:b33ibw5) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/tests/SIL.Machine.AspNetCore.Tests/Services/BiblicalRangeStringParserTests.cs#L73)):*
> ```Smalltalk
>         yield return new TestCaseData("NT,OT,-MRK,-EXO", new Dictionary<string, List<int>>(), true);
>         yield return new TestCaseData("OT,MAT1", new Dictionary<string, List<int>>(), true);
>         yield return new TestCaseData("OT,MAT-LUK", new Dictionary<string, List<int>>(), true);
> ```

Is this the same list from python?

johnml1135:
Hmmm.  It could make errors easier to handle.  Wouldn't it also be good then to pull textId filtering into Serval as well?

Enkidu93 (commented):
No, I think the idea is to keep the filtering in Machine, just move the actual parsing of the string to Machine. Ultimately, there will still be filtration code in all engine jobs regardless - it just means that Serval is responsible for parsing the range - which kind of makes sense semantically. This way too, we can catch that parsing error earlier on. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np6M_lF8vsP8iJ8I0vu:b-y78o3a)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`tests/SIL.Machine.AspNetCore.Tests/Services/BiblicalRangeStringParserTests.cs` line 73 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np6H4zi0IbM8UKzxHbZ:-Np6Jxv-1FY6RywSOhuY:b9idqdq) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/tests/SIL.Machine.AspNetCore.Tests/Services/BiblicalRangeStringParserTests.cs#L73)):*

Is this the same list from python?
</blockquote></details>

Yep, all the same test cases
___
*[`tests/SIL.Machine.AspNetCore.Tests/Services/data/paratext2/TermRenderings.xml` line 2 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np6HgwcD5wHTcwGIxV2:-Np6K-Ou4mmd4_HZyG5A:b-6rlabx) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/tests/SIL.Machine.AspNetCore.Tests/Services/data/paratext2/TermRenderings.xml#L2)):*

What is the purpose of these term renderings and other files?
</blockquote></details>

Originally, our only Paratext `NmtPreprocessBuildJob` tests were mapping a single Paratext project as source and target. That didn't allow me enough control to make good biblical range tests.

Enkidu93 (commented):
@ddaspit I'm realizing the downside of this approach is that we don't have access to the versification in Serval (which the `ScriptureRangeParser` is using to determine books/counts accurately). I'm not sure how we'd get around that unless you don't imagine that will really be an issue. Let me know as soon as possible what you think.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np6VreD5R11a6Xa06l6:b-wbwn35)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit and @johnml1135)

ddaspit (commented):
I had to write code that accesses the versification for the USFM generation PR. I will abstract that functionality into a service that you can use.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np6YMTLEpXr6tkCo-ap:b-2s44q5)* status: all files reviewed, 7 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

Enkidu93 (commented):
OK, I guess if you're interdependent, you might as well go all the way haha. Thank you Let me know when that's in place.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Np6ad974dOEepwxzo2g:bsct7rn)* status: all files reviewed, 7 unresolved discussions (waiting on @ddaspit and @johnml1135)

johnml1135:
*[`src/SIL.Machine.AspNetCore/Models/Corpus.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj2UahEtf_ohUiRWkv:-Npj3QFG7jemxz5VSrIE:b43nzdt) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Models/Corpus.cs#L11)):*

Chapters or ScriptureRange?
</blockquote></details>

I see Damien's comment at the bottom.  Ok.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 156 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj5vPTDlQ5TNBb6iYu:-Npj5vPTDlQ5TNBb6iYv:btkkd3o) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L156)):*
> ```Smalltalk
>                                             || (rcbp.Value.Count() > 0 && tobrc.Value.Count() == 0) //Empty list means all chapters from book
>                                     )
>                                     .Any(b => b);
> ```

That logic appears solid, namely:
* Both train on chapters and rows are dictionaries: book (string): chapters (list of ints)
* Join the two, looking for intersections.  If there is an intersection, well and good.  If, not, if both row and filter have a book, but the filter has no chapter, it is valid as well.  Any matches, include the verse.

The only thing I wonder is if could be optimized more (see here for intersection explanation - https://stackoverflow.com/questions/14527595/intersection-of-two-sets-in-most-optimized-way).  An alternative could be:
* Per row verse reference, get filter's book.  if ref chapter in filter list, good, or if filter list size == 0.  Any hits and it is included.  
* This would be O(# of verse references in row * number of chapters in book * rows)
* We could be crazy about optimization and make an array 256 boolean for the filter and then check the chapter as a lookup in that array and get O(# of verse references in row  * rows), but it would be more obtuse and give diminishing returns.

It may save some cycles - what do you think?  I am fine with it as is as I am not anticipating a meaningful performance hit even with O(N*M*P).

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 183 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj9SvHAgjnSUCjt7jc:-Npj9SvHAgjnSUCjt7jd:bntmtq) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L183)):*
> ```Smalltalk
>                             corpus.PretranslateAll
>                             || corpus.PretranslateTextIds.Contains(row.TextId)
>                             || isInPretranslateChapters
> ```

Just to clarify how references work.  There can be multiple references - is it the source and target reference?  Are we assuming only one actual verse reference?  And therefore only one book (Daniel vs. Bell and the Dragon)?  It's a severe edge case - just checking.

johnml1135 (commented):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @Enkidu93)

ddaspit (changes requested):
> Reviewed 8 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Scripture/ScriptureRangeParser.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-NpumPLy8Cg9Zv7wjKLn:-NpumPLy8Cg9Zv7wjKLo:b-6gd0of) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine/Scripture/ScriptureRangeParser.cs#L8)):*
> ```Smalltalk
> using SIL.Scripture;
> 
> public class ScriptureRangeParser
> ```

It would be nice to add a static convenience method `GetChapters(string chapterSelection, ScrVers versification = null)`.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 138 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-NpuuHVK9E-lMb7NFc9N:-NpuuHVK9E-lMb7NFc9O:b72mwmq) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L138)):*
> ```Smalltalk
>                     {
>                         Dictionary<string, List<int>>? rowChaptersPerBook = null;
>                         if (corpus.TrainOnChapters != null || corpus.PretranslateChapters != null)
> ```

This seems much simpler and clearer to me:

```
bool IsInChapters(Dictionary<string, List<int>> bookChapters, object rowRef)
{
    if (rowRef is not VerseRef vr)
        return false;
    return bookChapters.TryGetValue(vr.Book, out List<int>? chapters)
        && chapters.Contains(vr.ChapterNum);
}

if (corpus.TrainOnChapters is not null)
    isInTrainOnChapters = row.Refs.Any(r => IsInChapters(corpus.TrainOnChapters, r));
if (corpus.PretranslateChapters is not null)
    isInPretranslateChapters = row.Refs.Any(r => IsInChapters(corpus.PretranslateChapters, r));
```
Also, it would be slightly more efficient if`TrainOnChapters` and `PretranslateChapters` were `Dictionary<string, HashSet<int>>`.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 137 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np674nd5kyKjZkcg6ZV:-Npv0lx50BHzJeH8LD3_:b-dzlisu) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L137)):*

We should probably only do these checks if there is a biblical range - if not, just skip it.  Less computation?
</blockquote></details>

Yeah - and that should be rolled in with the refactoring of the logic Damien suggests. 
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 142 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np64ZI-CXnYeT_OAl1H:-Npv-jwW-UhIDC2Ff7Dc:b-t7f781) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L142)):*

So, to be clear, if the verse is in either the source or the target versification, it will work?  Is that what we want?
</blockquote></details>

I'm not sure what you mean here. Are you saying we should just use one of either SourceRefs or TargetRefs here (whichever we use Serval-side)? The refs is kind of a shorthand for SourceRefs if there are any; otehrwise, targetrefs. 
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 138 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-NpuuHVK9E-lMb7NFc9N:-Npuz2jS1MRASwdmstGf:b-dgd23z) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L138)):*

This seems much simpler and clearer to me:

```
bool IsInChapters(Dictionary<string, List<int>> bookChapters, object rowRef)
{
    if (rowRef is not VerseRef vr)
        return false;
    return bookChapters.TryGetValue(vr.Book, out List<int>? chapters)
        && chapters.Contains(vr.ChapterNum);
}

if (corpus.TrainOnChapters is not null)
    isInTrainOnChapters = row.Refs.Any(r => IsInChapters(corpus.TrainOnChapters, r));
if (corpus.PretranslateChapters is not null)
    isInPretranslateChapters = row.Refs.Any(r => IsInChapters(corpus.PretranslateChapters, r));
```
Also, it would be slightly more efficient if`TrainOnChapters` and `PretranslateChapters` were `Dictionary<string, HashSet<int>>`.
</blockquote></details>

Good call. This can be collapsed in a way that was more difficult before when the parsing was here. My mind is working at flu speed at the moment, but I don't think your code would properly handle the 'empty list = all chapters' idiom, right? It's an easy addition, but it just brought to my attention that we should make sure we're all on the same page about that. 

As for the HashSet - definitely, I can convert it to a hash set when it's being mapped (which is a cost in and of itself), but probably still cleaner. 
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 156 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj5vPTDlQ5TNBb6iYu:-Npv2DcW8eQO0qYpxp4U:b-vyjyep) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L156)):*

That logic appears solid, namely:
* Both train on chapters and rows are dictionaries: book (string): chapters (list of ints)
* Join the two, looking for intersections.  If there is an intersection, well and good.  If, not, if both row and filter have a book, but the filter has no chapter, it is valid as well.  Any matches, include the verse.

The only thing I wonder is if could be optimized more (see here for intersection explanation - https://stackoverflow.com/questions/14527595/intersection-of-two-sets-in-most-optimized-way).  An alternative could be:
* Per row verse reference, get filter's book.  if ref chapter in filter list, good, or if filter list size == 0.  Any hits and it is included.  
* This would be O(# of verse references in row * number of chapters in book * rows)
* We could be crazy about optimization and make an array 256 boolean for the filter and then check the chapter as a lookup in that array and get O(# of verse references in row  * rows), but it would be more obtuse and give diminishing returns.

It may save some cycles - what do you think?  I am fine with it as is as I am not anticipating a meaningful performance hit even with O(N*M*P).
</blockquote></details>

Definitely - the optimizations are endless haha. Because this is bounded (we know how many chapters there are), I don't know how crazy we need to get here. Let me know if you'd like a motion in this direction.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 183 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj9SvHAgjnSUCjt7jc:-Npv2fTX6vi309Umb4gI:b5k76rm) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L183)):*

Just to clarify how references work.  There can be multiple references - is it the source and target reference?  Are we assuming only one actual verse reference?  And therefore only one book (Daniel vs. Bell and the Dragon)?  It's a severe edge case - just checking.
</blockquote></details>

No, not source and target. If you peek at `ParallelTextRow`, you'll see that the refs is shorthand for source refs if there are any; otherwise, target refs. For the most part, there should only be one verse reference. Damien can correct me, but I think the idea of multiple references is more that sometimes a chunk of verses is translated as a unit because of the peculiarities of a language (like say it's topic-comment and rearranging phrases from contiguous verses is necessary to make the reading natural). I could be way off haha. @ddaspit ?

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 156 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj5vPTDlQ5TNBb6iYu:-Npv6d2hAr3s9vEJBybm:bbbjo80) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L156)):*

Definitely - the optimizations are endless haha. Because this is bounded (we know how many chapters there are), I don't know how crazy we need to get here. Let me know if you'd like a motion in this direction.
</blockquote></details>

I think Damien resolves all optimization issues, including the hashset - bringing it to O(n).

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 142 at r1](https://reviewable.io/reviews/sillsdev/machine/161#-Np64ZI-CXnYeT_OAl1H:-Npv75_n4BRbHALFYbUl:b-jvd9t0) ([raw file](https://github.com/sillsdev/machine/blob/307eba72d68474a187ef1d68f2bfb1f1d81923ef/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L142)):*

I'm not sure what you mean here. Are you saying we should just use one of either SourceRefs or TargetRefs here (whichever we use Serval-side)? The refs is kind of a shorthand for SourceRefs if there are any; otehrwise, targetrefs.
</blockquote></details>

It should be good enough.  If we run into future problems, we can seek to address it.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 183 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-Npj9SvHAgjnSUCjt7jc:-NpvO1oU03vUdNufCGkF:b-p11qga) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L183)):*

No, not source and target. If you peek at `ParallelTextRow`, you'll see that the refs is shorthand for source refs if there are any; otherwise, target refs. For the most part, there should only be one verse reference. Damien can correct me, but I think the idea of multiple references is more that sometimes a chunk of verses is translated as a unit because of the peculiarities of a language (like say it's topic-comment and rearranging phrases from contiguous verses is necessary to make the reading natural). I could be way off haha. @ddaspit ?
</blockquote></details>

@Enkidu93, you are correct. That is an accurate description.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 142 at r3](https://reviewable.io/reviews/sillsdev/machine/161#-NpvMQuRBhBnnB6dRViP:-NpvMQuRBhBnnB6dRViQ:b-o5c0gq) ([raw file](https://github.com/sillsdev/machine/blob/4632deef84f0fae2df661b4ff599e5782839b1e1/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L142)):*
> ```Smalltalk
>                                 return false;
>                             return bookChapters.TryGetValue(vr.Book, out HashSet<int>? chapters)
>                                 && (chapters.Contains(vr.ChapterNum) || chapters.Count() == 0);
> ```

You should use the `Count` property instead of the `Count()` LINQ method. Using `Count()` can be dangerous, because in certain situations it computes the count by enumerating over the whole collection.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: 14 of 15 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Scripture/ScriptureRangeParser.cs` line 8 at r2](https://reviewable.io/reviews/sillsdev/machine/161#-NpumPLy8Cg9Zv7wjKLn:-NpvGXjX9KgqJKHnR09P:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/1119992163bcc905b388a499e432f1ff7840e896/src/SIL.Machine/Scripture/ScriptureRangeParser.cs#L8)):*

It would be nice to add a static convenience method `GetChapters(string chapterSelection, ScrVers versification = null)`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 142 at r3](https://reviewable.io/reviews/sillsdev/machine/161#-NpvMQuRBhBnnB6dRViP:-NpvVwau6R5HGtw8aXbI:b-9fezxa) ([raw file](https://github.com/sillsdev/machine/blob/4632deef84f0fae2df661b4ff599e5782839b1e1/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L142)):*

You should use the `Count` property instead of the `Count()` LINQ method. Using `Count()` can be dangerous, because in certain situations it computes the count by enumerating over the whole collection.
</blockquote></details>

Done. Thank you! Good catch. I'm so accustomed to using the method because I'm working with a generic IEnumerable.

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161#-:-Npz29ZE48WMCfo0rCOQ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

johnml1135 (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/161)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #164, opened 2024-02-05
Nmt download

johnml1135:
Fixes https://github.com/sillsdev/serval/issues/268

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/164)


johnml1135:
*[`src/SIL.Machine/Translation/ModelDownloadUrl.cs` line  at r2](https://reviewable.io/reviews/sillsdev/machine/164#-Npzs21NEjBy97EVV-zc:-Nq-0Coh7CcnaLiJoYF8:bb6un1j) ([raw file](https://github.com/sillsdev/machine/blob/77a7f9b43428d49de762136e8124d582879fde29/src/SIL.Machine/Translation/ModelDownloadUrl.cs#L)):*

This should be moved to `SIL.Machine.AspNetCore/Models`.
</blockquote></details>

Done

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 211 at r2](https://reviewable.io/reviews/sillsdev/machine/164#-NpzqZ5Q82qkEGoRjzHJ:-Nq-0MQZ0WG1tws41656:b8tcxo1) ([raw file](https://github.com/sillsdev/machine/blob/11d098a8f6b2239964e5d9be100525a35fa137eb/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L211)):*

I like `ToArray` better. It is clearer.
</blockquote></details>

Quick fix likes the collection expression, but I can change it.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 421 at r2](https://reviewable.io/reviews/sillsdev/machine/164#-NpzrX9b1AB-AAR1tGUW:-Nq-1b6wEg78bnxdNo4Y:bxbmhzg) ([raw file](https://github.com/sillsdev/machine/blob/77a7f9b43428d49de762136e8124d582879fde29/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L421)):*

This should be run using a `BackgroundService`. See `RecurrentTask`.  `ClearMLMonitorService` and `SmtTransferEngineCommitService` are existing examples.
</blockquote></details>

ok - that makes sense

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/164)* status: 3 of 23 files reviewed, 4 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine.AspNetCore/Models/TranslationEngine.cs` line 10 at r2](https://reviewable.io/reviews/sillsdev/machine/164#-NpzrBGfBQ0D5vzir0N_:-Nq-4_uN7hYGUNMAK1kK:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/77a7f9b43428d49de762136e8124d582879fde29/src/SIL.Machine.AspNetCore/Models/TranslationEngine.cs#L10)):*

This should be `IsModelPersisted`.
</blockquote></details>

Done.

ddaspit (changes requested):
> Reviewed 6 of 21 files at r1, 2 of 13 files at r2, 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/164)* status: all files reviewed, 18 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 410 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Z44FC-0Y0wlcRvsN:-Nq-Z44FC-0Y0wlcRvsO:bmb2hla) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L410)):*
> ```Smalltalk
>     public static IMachineBuilder AddModelCleanupService(this IMachineBuilder builder)
>     {
>         builder.Services.AddSingleton<CleanupOldModelsService>();
> ```

You only need to add a hosted service as a singleton if you want to allow it to be injected into other services.
___
*[`src/SIL.Machine.AspNetCore/Services/IFileStorage.cs` line 17 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-TfGD2qAZkl3T5lAl:-Nq-TfGD2qAZkl3T5lAm:b-9csayb) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/IFileStorage.cs#L17)):*
> ```Smalltalk
>     Task<Stream> OpenWriteAsync(string path, CancellationToken cancellationToken = default);
> 
>     Task<string> GetPresignedUrlAsync(string path, int minutesToExpire, CancellationToken cancellationToken = default);
> ```

This should be named `GetDownloadUrlAsync`.
___
*[`src/SIL.Machine.AspNetCore/Services/InMemoryStorage.cs` line 105 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-V074D44S15FxPZXL:-Nq-V074D44S15FxPZXM:b-ki7aas) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/InMemoryStorage.cs#L105)):*
> ```Smalltalk
>     )
>     {
>         return Task.FromResult(path);
> ```

We should throw `NotSupportedException`.
___
*[`src/SIL.Machine.AspNetCore/Services/ISharedFileService.cs` line 5 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-X6CbCdTC4ssWFw4Q:-Nq-X6CbCdTC4ssWFw4R:bymmahv) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/ISharedFileService.cs#L5)):*
> ```Smalltalk
> public interface ISharedFileService
> {
>     public const string ModelDirectory = "models/";
> ```

This constant should be moved to `NmtEngineService`.
___
*[`src/SIL.Machine.AspNetCore/Services/LocalStorage.cs` line 45 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-VoszEdYCpjrPFpg-:-Nq-VoszEdYCpjrPFpg0:b-ki7aas) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/LocalStorage.cs#L45)):*
> ```Smalltalk
>     )
>     {
>         return Task.FromResult(path);
> ```

We should throw `NotSupportedException`.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtClearMLBuildJobFactory.cs` line 47 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-ZYPbDKQY-WklhnEM:-Nq-ZYPc5AzLWOwDn7T_:b-jigni9) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtClearMLBuildJobFactory.cs#L47)):*
> ```Smalltalk
>                 + $"    'shared_file_folder': '{folder}',\n"
>                 + (buildOptions is not null ? $"    'build_options': '''{buildOptions}''',\n" : "")
>                 + (engine.IsModelPersisted ? $"    'save_model': '{engineId}_{engine.BuildRevision + 1}',\n" : "")
> ```

Why `+ 1`? Aren't we getting the current build revision?
___
*[`src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs` line 124 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-WtS85f_kXGxwQ8tf:-Nq-WtS85f_kXGxwQ8tg:be5cote) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs#L124)):*
> ```Smalltalk
>         TranslationEngine engine = await GetEngineAsync(engineId, cancellationToken);
>         if (!engine.IsModelPersisted)
>             throw new InvalidOperationException(
> ```

You should throw a `NotSupportedException`.
___
*[`src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs` line 148 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Wba0A7fHBKN5n24a:-Nq-Wba0A7fHBKN5n24b:b-5ztuzv) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs#L148)):*
> ```Smalltalk
>             ).ToString(),
>             ModelRevision = engine.BuildRevision,
>             ExipiresAt = DateTime.UtcNow.AddMinutes(MinutesToExpire)
> ```

We should either pass in the `DateTime` into the `_sharedFileService.GetPresignedUrlAsync` call or return the expiration time from the method.
___
*[`src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs` line 146 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-XiH8DoJkVFjXPEAC:-Nq-XiH8DoJkVFjXPEAD:b-plovif) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs#L146)):*
> ```Smalltalk
>             throw new RpcException(new Status(StatusCode.Aborted, e.Message));
>         }
>         catch (FileNotFoundException e)
> ```

We shouldn't handle this exception. It only happens if there is something really wrong.
___
*[`src/SIL.Machine.AspNetCore/Services/SharedFileService.cs` line 62 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Y5XU7VS8Vx-mJlnv:-Nq-Y5XU7VS8Vx-mJlnw:b-gznurh) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SharedFileService.cs#L62)):*
> ```Smalltalk
>         string presignedUrl = path;
>         if (_baseUri is not null)
>             if (_baseUri.Scheme == "s3")
> ```

Why are we checking if this is an S3 file storage here? The `IFileStorage` abstraction should hide any internal implementation details like this.
___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-Nq-YSxJ3tVHUpW1d-a-:bo4pxit) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*
> ```Smalltalk
>                 SourceLanguage = sourceLanguage,
>                 TargetLanguage = targetLanguage,
>                 IsModelPersisted = isModelPersisted
> ```

SMT models are always persisted.
___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 223 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YXIr7fr7xCd-Xh-M:-Nq-YXIr7fr7xCd-Xh-N:bs8vcpb) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L223)):*
> ```Smalltalk
>     )
>     {
>         throw new NotImplementedException();
> ```

This should be `NotSupportedException`.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 3 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Ym05-ytM42-Zx8t4:-Nq-Ym05-ytM42-Zx8t5:bsgmwkc) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L3)):*
> ```Smalltalk
> namespace SIL.Machine.AspNetCore.Services;
> 
> public class CleanupOldModelsService(
> ```

This should be named `ModelCleanupService`.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YzyRELBzUoOtPnaj:-Nq-YzyRELBzUoOtPnak:b91o2wd) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L10)):*
> ```Smalltalk
> ) : RecurrentTask("Cleanup Old Models Service", services, RefreshPeriod, logger)
> {
>     public ISharedFileService SharedFileService { get; } = sharedFileService;
> ```

Is this `public` for a reason?
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_Dq_8dgwIsyNolhR:-Nq-_Dq_8dgwIsyNolhS:b-3kvqpo) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L19)):*
> ```Smalltalk
>     protected override async Task DoWorkAsync(IServiceScope scope, CancellationToken cancellationToken)
>     {
>         await CheckModelsAsync();
> ```

You should pass in the cancellation token and use it for all async calls.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 22 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Yvnz6xJuigNAKdjg:-Nq-Yvnz6xJuigNAKdjh:b91o2wd) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L22)):*
> ```Smalltalk
>     }
> 
>     public async Task CheckModelsAsync()
> ```

Is this `public` for a reason?
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-U08i90SdRTnu_5Dh:-Nq-U08i90SdRTnu_5Di:b-tdr2v8) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L25)):*
> ```Smalltalk
>     {
>         _logger.LogInformation("Running model cleanup job");
>         var paths = await SharedFileService.ListFilesAsync(ISharedFileService.ModelDirectory);
> ```

Please try to remember to only use `var` when the type is explicit elsewhere in the line.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq-_noC39TAEu2RwMka:brfv43e) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*
> ```Smalltalk
>     {
>         // If a file has been requested to be deleted twice, delete it. Otherwise, mark it for deletion.
>         if (_filesPreviouslyMarkedForDeletion.Contains(filename))
> ```

We need a more deterministic method for cleaning up old models that is dependent on temporary state that gets wiped out when the server restarts. For example, we could delete models 30 days after they were created.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 410 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Z44FC-0Y0wlcRvsN:-Nq04LGW1hiIxDpRv0pM:bjfydt0) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L410)):*

You only need to add a hosted service as a singleton if you want to allow it to be injected into other services.
</blockquote></details>

Ah - forgot.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 410 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Z44FC-0Y0wlcRvsN:-Nq05GKW5Df7Udz2qpLc:b-8jghz8) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L410)):*

Ah - forgot.
</blockquote></details>

Hmm.  If I remove Hosted service, it doesn't get called.  I believe I do want it as a hosted service: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-8.0&tabs=visual-studio

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YzyRELBzUoOtPnaj:-Nq05lNz8amYzpiZBHxa:b2nhky0) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L10)):*

Is this `public` for a reason?
</blockquote></details>

No.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_Dq_8dgwIsyNolhR:-Nq07B2j9V78QUYverWd:bb6un1j) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L19)):*

You should pass in the cancellation token and use it for all async calls.
</blockquote></details>

Done

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 22 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Yvnz6xJuigNAKdjg:-Nq07Dpq21utHoM6Ypee:by1l15q) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L22)):*

Is this `public` for a reason?
</blockquote></details>

I was wanting to test it.  I was able to find a work around as per here: https://stackoverflow.com/questions/9122708/unit-testing-private-methods-in-c-sharp.  I could have also used "InternalsVisibleTo", but was unsure the best way to approach it.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 25 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-U08i90SdRTnu_5Dh:-Nq0BRJY14xWrPXhgSLc:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L25)):*

Please try to remember to only use `var` when the type is explicit elsewhere in the line.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq0BWQl1eofunYK3s9l:boug5qw) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*

We need a more deterministic method for cleaning up old models that is dependent on temporary state that gets wiped out when the server restarts. For example, we could delete models 30 days after they were created.
</blockquote></details>

Any filename that does not correspond to a database entry gets wiped out.  I think that is strong enough.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/IFileStorage.cs` line 17 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-TfGD2qAZkl3T5lAl:-Nq0HaWV2_V45EsXHaX6:b7rf7y6) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/IFileStorage.cs#L17)):*

This should be named `GetDownloadUrlAsync`.
</blockquote></details>

I thought I had gotten rid of all of them.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ISharedFileService.cs` line 5 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-X6CbCdTC4ssWFw4Q:-Nq0NcBD4H6UvcVP7bqz:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/ISharedFileService.cs#L5)):*

This constant should be moved to `NmtEngineService`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtClearMLBuildJobFactory.cs` line 47 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-ZYPbDKQY-WklhnEM:-Nq0Neth54PTjr0awq1v:bq8fgaf) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtClearMLBuildJobFactory.cs#L47)):*

Why `+ 1`? Aren't we getting the current build revision?
</blockquote></details>

Added comment for clarity.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs` line 124 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-WtS85f_kXGxwQ8tf:-Nq0OFHT7AhKAWnbNeSE:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs#L124)):*

You should throw a `NotSupportedException`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs` line 148 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Wba0A7fHBKN5n24a:-Nq0Psnl4H2eYu6vC_XN:brt44bj) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/NmtEngineService.cs#L148)):*

We should either pass in the `DateTime` into the `_sharedFileService.GetPresignedUrlAsync` call or return the expiration time from the method.
</blockquote></details>

You're right - I knew it was funky.  I will pass it down.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs` line 146 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-XiH8DoJkVFjXPEAC:-Nq0Q7Jb6lSSNJBK6Xmr:b-e4miid) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs#L146)):*

We shouldn't handle this exception. It only happens if there is something really wrong.
</blockquote></details>

Sure - a 500 code is appropriate.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SharedFileService.cs` line 62 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Y5XU7VS8Vx-mJlnv:-Nq0QCViAaF4Qnamt_qJ:b-b7rruk) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SharedFileService.cs#L62)):*

Why are we checking if this is an S3 file storage here? The `IFileStorage` abstraction should hide any internal implementation details like this.
</blockquote></details>

I don't know - or why I chose the Uri format.  Moved everything to strings.  I think I was thinking that we would want an implementation for inmemory or filestorage.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-Nq0SUGLA7CK8vEk8a_Y:b-gd8kxu) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

SMT models are always persisted.
</blockquote></details>

Ok - hardcode it?

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 223 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YXIr7fr7xCd-Xh-M:-Nq0SbsZ7M53SHxPPS2c:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L223)):*

This should be `NotSupportedException`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq2iRrZAvKlh7SyhIak:b-d0mzca) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*

Any filename that does not correspond to a database entry gets wiped out.  I think that is strong enough.
</blockquote></details>

I will update this logic to be dead reckoning.  We know what filenames should exist - let's just delete any that aren't those.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/InMemoryStorage.cs` line 105 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-V074D44S15FxPZXL:-Nq0IIYw2lM4cFfY5_tA:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/InMemoryStorage.cs#L105)):*

We should throw `NotSupportedException`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 3 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Ym05-ytM42-Zx8t4:-Nq05T1d7vJjsb1n_NPU:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L3)):*

This should be named `ModelCleanupService`.
</blockquote></details>

Done.

ddaspit (commented):
> Reviewed 17 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/164)* status: all files reviewed, 9 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 410 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Z44FC-0Y0wlcRvsN:-Nq3zwWUAnzjSsindgz1:bpe6o42) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L410)):*

Hmm.  If I remove Hosted service, it doesn't get called.  I believe I do want it as a hosted service: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-8.0&tabs=visual-studio
</blockquote></details>

All you should need is
```
builder.Services.AddHostedService<ModelCleanupService>();
```
___
*[`src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs` line 149 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq437Sj5UfqTJUVOrcT:-Nq437Sk4T_Fou-h62Cp:bbpel6s) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ServalTranslationEngineServiceV1.cs#L149)):*
> ```Smalltalk
>             };
>         }
>         catch (InvalidOperationException e)
> ```

I don't think we need to catch `InvalidOperationException` anymore.
___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-Nq45LTB-cAAZbafE8GY:b4a4ncm) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

Ok - hardcode it?
</blockquote></details>

That works for Machine, but what about Serval? We should make `IsModelPersisted` optional. We can throw an exception if they pass an invalid value.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YzyRELBzUoOtPnaj:-Nq43ffp8zES6itviZgf:b-c9quxy) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L10)):*

No.
</blockquote></details>

Now that it is private, this should be a field instead of a property.
___
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq43sQbEkGj_nsfDEh1:b-ow3egu) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*

I will update this logic to be dead reckoning.  We know what filenames should exist - let's just delete any that aren't those.
</blockquote></details>

I'm not sure I understand the purpose of marking it for deletion. Why not just delete it?
___
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 15 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq440qt-prBNXngo9Cc:-Nq440qt-prBNXngo9Cd:brxby2j) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L15)):*
> ```Smalltalk
>     private List<string> _filesPreviouslyMarkedForDeletion = [];
>     private readonly List<string> _filesNewlyMarkedForDeletion = [];
>     private static readonly TimeSpan RefreshPeriod = TimeSpan.FromSeconds(10);
> ```

This seems very short to me.
___
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 25 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44Ret4FJxg3WXe5X-:-Nq44Ret4FJxg3WXe5X0:byjf0q) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L25)):*
> ```Smalltalk
>     {
>         _logger.LogInformation("Running model cleanup job");
>         IReadOnlyCollection<string> paths = await SharedFileService.ListFilesAsync(
> ```

I really don't love that we enumerate all model files and all engines in order to check for which models to delete. Is there any way we can design this so that we avoid these expensive operations?
___
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 31 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44BEr9YOsfYjJtanE:-Nq44BEr9YOsfYjJtanF:b-knbsgx) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L31)):*
> ```Smalltalk
>         // Get all engine ids from the database
>         Dictionary<string, int> engineIdsToRevision = _engines
>             .GetAllAsync(cancellationToken: cancellationToken)
> ```

You should await this call.
___
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 32 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44DlV8xLZBR-SZ0KR:-Nq44DlV8xLZBR-SZ0KS:b1oykc) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L32)):*
> ```Smalltalk
>         Dictionary<string, int> engineIdsToRevision = _engines
>             .GetAllAsync(cancellationToken: cancellationToken)
>             .Result.Select(e => (e.EngineId, e.BuildRevision))
> ```

There is no need for separate `Select` and `ToDictionary` calls. You can specify the key and value in the `ToDictionary` call.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs` line 410 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-Z44FC-0Y0wlcRvsN:-Nq4VhIc8mQdbMXOy17s:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Configuration/IMachineBuilderExtensions.cs#L410)):*

All you should need is
```
builder.Services.AddHostedService<ModelCleanupService>();
```
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 10 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YzyRELBzUoOtPnaj:-Nq4VyK6BrPbZWcBfuGW:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L10)):*

Now that it is private, this should be a field instead of a property.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 15 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq440qt-prBNXngo9Cc:-Nq4W-AcE6DHITG5yVg2:b-uop5vx) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L15)):*

This seems very short to me.
</blockquote></details>

This was used for testing - I was supposed to change it back to 1 day.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 25 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44Ret4FJxg3WXe5X-:-Nq4W9Ku4wzOB2WM2m9a:b-q30dqf) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L25)):*

I really don't love that we enumerate all model files and all engines in order to check for which models to delete. Is there any way we can design this so that we avoid these expensive operations?
</blockquote></details>

The engines are not keys on engineId's, therefore every engine lookup is a O(n) operation.  Therefore, if each engine had a model, the opearation would be O(n*n).  By listing them out once, first, we can reduce this burden.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 31 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44BEr9YOsfYjJtanE:-Nq4Z0uR5BLkyTHP-65Q:b-oqukpj) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L31)):*

You should await this call.
</blockquote></details>

Reworked for clarity.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 32 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44DlV8xLZBR-SZ0KR:-Nq4ZHBs0z6WeDKhFFA7:br10od) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L32)):*

There is no need for separate `Select` and `ToDictionary` calls. You can specify the key and value in the `ToDictionary` call.
</blockquote></details>

The wonders of dotnet...

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq4ZMK06yzE2cSTM4s8:bntni0o) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*

I'm not sure I understand the purpose of marking it for deletion. Why not just delete it?
</blockquote></details>

If the file is actively being downloaded...

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-Nq4_jtaARvuBGFP004X:b-bw16s9) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

That works for Machine, but what about Serval? We should make `IsModelPersisted` optional. We can throw an exception if they pass an invalid value.
</blockquote></details>

Is invalid false, or anything other than null?

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 25 at r4](https://reviewable.io/reviews/sillsdev/machine/164#-Nq44Ret4FJxg3WXe5X-:-Nq8QucHEjwZOuMqaYiP:b-si8zen) ([raw file](https://github.com/sillsdev/machine/blob/9dd74a680cb096b93d8ddf1a628f13d461eb155b/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L25)):*

I can live with it, but there is an alternative. If we tracked the model files in Mongo, we could retrieve just the files that need to be deleted instead of retrieving all of the model files and all of the engines.
</blockquote></details>

I think this is a simple and good (enough) path forward right now and am inclined just to move forward with it.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-_noC39TAEu2RwMk_:-Nq8RAEd3CpaoWmJ_iKG:b-ft25as) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/CleanupOldModelsService.cs#L68)):*

It is still not clear to me what this is buying us. I also don't like relying on temporary state to cleanup the files. I think we should simplify the logic and just delete the file.
</blockquote></details>

The biggest thing is deleting a model as it is being downloaded and then a build finishes, the download will crash ungraciously.  If this is ok, then we can delete right away.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-Nq8VTxK2l5wMGjT5EvF:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

We should throw an exception if `false`. `null` or `true` is fine.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Models/TranslationEngine.cs` line 10 at r5](https://reviewable.io/reviews/sillsdev/machine/164#-Nq8H-h20nRmTnlrCfd4:-Nq8r2AlAUIabNQH0MBJ:bt2vr1b) ([raw file](https://github.com/sillsdev/machine/blob/ad13e4448f774b32a38d78922e0c7cf757d62e55/src/SIL.Machine.AspNetCore/Models/TranslationEngine.cs#L10)):*

This does not need to be nullable. It will always be set.
</blockquote></details>

Fixed.

ddaspit (commented):
> Reviewed 3 of 6 files at r6, 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/164)* status: all files reviewed, 2 unresolved discussions (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-NqDsl_z10LNdXJ2HT3o:b-6o8hik) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

Done.
</blockquote></details>

I don't see where the exception is thrown. The SMT engine doesn't currently support `IsModelPersisted == false`, so we should throw an `InvalidOperationException` that gets caught in `ServalTranslationEngineServiceV1` and rethrown as an `RpcException` with the status code `InvalidArgument`.
___
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 13 at r7](https://reviewable.io/reviews/sillsdev/machine/164#-NqDrvnv72K-mBunblGx:-NqDrvnw5sDnDZvOAEbN:bl5dh7) ([raw file](https://github.com/sillsdev/machine/blob/5fe8bcc97f218cc440791a3defb7b07b371df750/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L13)):*
> ```Smalltalk
>     private ILogger<ModelCleanupService> _logger = logger;
>     private IRepository<TranslationEngine> _engines = engines;
>     private List<string> _filesPreviouslyMarkedForDeletion = [];
> ```

`__filesPreviouslyMarkedForDeletion` and `_filesNewlyMarkedForDeletion` can be removed. The rest of the fields should be `readonly`.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs` line 13 at r7](https://reviewable.io/reviews/sillsdev/machine/164#-NqDrvnv72K-mBunblGx:-NqE2IJs0ZDxAz5KJi_o:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/5fe8bcc97f218cc440791a3defb7b07b371df750/src/SIL.Machine.AspNetCore/Services/ModelCleanupService.cs#L13)):*

`__filesPreviouslyMarkedForDeletion` and `_filesNewlyMarkedForDeletion` can be removed. The rest of the fields should be `readonly`.
</blockquote></details>

Done.

johnml1135:
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-NqE2Rf1DXrEt5FJvtK3:b4mrzy0) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

I don't see where the exception is thrown. The SMT engine doesn't currently support `IsModelPersisted == false`, so we should throw an `InvalidOperationException` that gets caught in `ServalTranslationEngineServiceV1` and rethrown as an `RpcException` with the status code `InvalidArgument`.
</blockquote></details>

10 lines above there is a "NotSupportedException".  What is the logic we should use for these two exceptions?  https://stackoverflow.com/questions/12669805/when-to-use-invalidoperationexception-or-notsupportedexception?

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/164#-:-NqEBek_0htCqxlNka8b:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

___
*[`src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs` line 46 at r3](https://reviewable.io/reviews/sillsdev/machine/164#-Nq-YSxJ3tVHUpW1d-_z:-NqEBL-d3MHYajU6Kh5h:bfpqp3t) ([raw file](https://github.com/sillsdev/machine/blob/8d6702d386ef808ca0589ea683ae58d240007389/src/SIL.Machine.AspNetCore/Services/SmtTransferEngineService.cs#L46)):*

10 lines above there is a "NotSupportedException".  What is the logic we should use for these two exceptions?  https://stackoverflow.com/questions/12669805/when-to-use-invalidoperationexception-or-notsupportedexception?
</blockquote></details>

I didn't see it. I think the `NotSupportedException` is sufficient for this case.

PR #165, opened 2024-02-06
Cancel build gRPC endpoint returns Aborted when no build is running

### LINKED ISSUE #162: CancelBuildAsync return StatusCode.Aborted if already cancelled
If I am correct, when calling the GRPC CancelBuildAsync twice, the second time it should throw an exception with StatusCode.Aborted.  We could also add a simple E2E test to verify that this works.  This should catch a race condition where a person cancels a build, but it's not really cancelled yet.  We should also handle starting a build before the previous is truly canceled.

@ddaspit - am I understanding this correctly?

ddaspit:
- fixes #162

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/165)


johnml1135 (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/165)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #170, opened 2024-02-13
Editorconfig

johnml1135:
Update everything!

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/170)


Enkidu93 (changes requested):
> Reviewed 139 of 355 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/170)* status: 139 of 355 files reviewed, 5 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Annotations/AnnotatedStringData.cs` line 85 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-Nqbe9RZ7Ck4iR6fWEmW:-Nqbe9RZ7Ck4iR6fWEmX:b-tq72bt) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Annotations/AnnotatedStringData.cs#L85)):*
> ```Smalltalk
>             foreach (Annotation<int> ann in toRemove)
>                 if (ann.List != null)
>                     ann.Remove(false);
> ```

Verify for me, John (similar to my comment in the serval PR): Would including braces here be flagged? I just wonder if this is really a good default. The lack of braces makes me uncomfortable. In this particular function (which is unlikely to be extended), I don't see an issue, but I wonder if this choice of whether or not to use braces is something we should leave up to the individual coding that particular bit of code (without flooding vscode with warnings or having to put in ignores everywhere).
___
*[`src/SIL.Machine/Annotations/Shape.cs` line 200 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbfTwuBzpx7zfL80E6:-NqbfTwuBzpx7zfL80E7:b58qq09) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Annotations/Shape.cs#L200)):*
> ```Smalltalk
>                         RelabelMinimumSparseEnclosingRange(curNode);
>                 }
>                 else if (curNode.Tag + 1 == curNode.Next.Tag)
> ```

Similar to above - it also reduces readability in my opinion.
___
*[`src/SIL.Machine/FiniteState/DeterministicFstTraversalInstance.cs` line 37 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbgljG-ukwKKScgw81:-NqbgljG-ukwKKScgw82:b-w99gwa) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/FiniteState/DeterministicFstTraversalInstance.cs#L37)):*
> ```Smalltalk
> 
>             var otherDfst = (DeterministicFstTraversalInstance<TData, TOffset>)other;
>             var outputMappings = Output
> ```

There are a few of these throughout: To me, the type is not perfectly clear, so I wouldn't use `var` here. How exactly does it decide when it's clear/unclear?
___
*[`src/SIL.Machine/Tokenization/LatinWordTokenizer.cs` line 134 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbitdN8jn-ixhwtIO-:-NqbitdN8jn-ixhwtIO0:b-o9u0vn) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Tokenization/LatinWordTokenizer.cs#L134)):*
> ```Smalltalk
>             }
>             else if (context.WordStart == -1)
>                 context.WordStart = context.Index;
> ```

This isn't an issue, really, but we should have some standard of what abbreviations in variable names are acceptable. Not worth switching it back, but I do think ctx or ctxt are very standard, unambiguous names; similarly, stddev elsewhere for standard deviation. It seems to me that as long as you have a clear name, shorter is better (e.g. col is better than column and num than number). Is the plan to just avoid abbreviations altogether?
___
*[`src/SIL.Machine/Translation/EcmScoreInfo.cs` line 36 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbjdcoA7sm-E4ZbI9X:-NqbjdcoA7sm-E4ZbI9Y:b-qdk73d) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Translation/EcmScoreInfo.cs#L36)):*
> ```Smalltalk
>         public int[] GetLastInsPrefixWordFromEsi()
>         {
>             int[] results = new int[Operations.Count];
> ```

Odd that the type is required here - like I said before, how exactly does it decide?

Enkidu93:
(About halfway through - I'll resume in a bit). Thank you for doing this John! Once the configuration is squared away, this will be very helpful.

johnml1135:
*[`src/SIL.Machine/FiniteState/DeterministicFstTraversalInstance.cs` line 37 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbgljG-ukwKKScgw81:-NqbrNrK1obwIxCxV4pq:b-svznze) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/FiniteState/DeterministicFstTraversalInstance.cs#L37)):*

There are a few of these throughout: To me, the type is not perfectly clear, so I wouldn't use `var` here. How exactly does it decide when it's clear/unclear?
</blockquote></details>

The ones I don't like as much are when you say "ToDictionary" or similar, it says the type is clear, but when the type is used as a type passed to the member, it's not clear.  I disagree with both.  The alternative is to not use the system...

johnml1135:
*[`src/SIL.Machine/Tokenization/LatinWordTokenizer.cs` line 134 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbitdN8jn-ixhwtIO-:-NqbscCA1TBryOQzxTNt:b-d8c1wl) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Tokenization/LatinWordTokenizer.cs#L134)):*

This isn't an issue, really, but we should have some standard of what abbreviations in variable names are acceptable. Not worth switching it back, but I do think ctx or ctxt are very standard, unambiguous names; similarly, stddev elsewhere for standard deviation. It seems to me that as long as you have a clear name, shorter is better (e.g. col is better than column and num than number). Is the plan to just avoid abbreviations altogether?
</blockquote></details>

It was hard to think about this one as well.  There is a time for abbreviation - but it should be limited and likely common to all the code.  `context` is used many times and should likely be spelled out each time.  For stddev, having it shorter is probably fine (didn't need to be changed).  I don't know if we should be changing every min, max, ave, etc.  In general, longer names for clarity usually appears to be preferred.

johnml1135:
*[`src/SIL.Machine/Translation/EcmScoreInfo.cs` line 36 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbjdcoA7sm-E4ZbI9X:-NqbtLFG3BmGIAwmtaq8:bvmitsr) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Translation/EcmScoreInfo.cs#L36)):*

Odd that the type is required here - like I said before, how exactly does it decide?
</blockquote></details>

`csharp_style_var_for_built_in_types = false`
Again, I don't fully like it but I don't know a better alternative.  Let's talk about these issues and see if we can come up with the best path forward.

johnml1135:
*[`src/SIL.Machine/Tokenization/LatinWordTokenizer.cs` line 134 at r1](https://reviewable.io/reviews/sillsdev/machine/170#-NqbitdN8jn-ixhwtIO-:-Nqbw7Ns1lNI2I3xxa6B:bcj3q9t) ([raw file](https://github.com/sillsdev/machine/blob/3435e68a316b5782dcda8450d18793ca50f27f28/src/SIL.Machine/Tokenization/LatinWordTokenizer.cs#L134)):*

It was hard to think about this one as well.  There is a time for abbreviation - but it should be limited and likely common to all the code.  `context` is used many times and should likely be spelled out each time.  For stddev, having it shorter is probably fine (didn't need to be changed).  I don't know if we should be changing every min, max, ave, etc.  In general, longer names for clarity usually appears to be preferred.
</blockquote></details>

We could use this guide: https://github.com/abbrcode/abbreviations-in-code

johnml1135:
I will start a new one

PR #176, opened 2024-02-29
Remove required property from optional ClearMLTask attributes

Enkidu93:
All of these properties are technically optional according to the docs. Should we make them all not required? It seems like we might want to avoid processing tasks that, for example, don't have an ID (And I'm not sure how that would ever happen anyways).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/176)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/176#-:-NrquGdSEIZLlMR8Q52k:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @johnml1135)

PR #179, opened 2024-04-02
Add support for non-verse text segments in Scripture corpora

ddaspit:
- add new ScriptureRef corpus ref class
- update Scripture corpora classes to use ScriptureRef
- add ScriptureRefUsfmParserHandlerBase class to track ScriptureRef in USFM
- update UsfmTextUpdater and UsfmTextBase to use ScriptureRefUsfmParserHandlerBase
- add support for updating non-Scripture paragraphs and notes
- update NmtPreprocessBuildJob to support non-Scripture segments

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)


Enkidu93 (changes requested):
Big update! Pretty cool to have this capability - thank you for your hard work :heart: . Not much I feel equipped to comment on - the rest is all on you, John :laughing: . 

> Reviewed 34 of 34 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179#-:-NuZdYbz-FvaWBlZjuhj:bxsar9f)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 42 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZ_L6j5k8h8i9pXu_o:-NuZ_L6j5k8h8i9pXu_p:b1v6nkw) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/ScriptureRef.cs#L42)):*
> ```Smalltalk
> 
>         public VerseRef VerseRef { get; }
>         public IReadOnlyList<ScriptureElement> Path { get; }
> ```

Why is this called `Path` given it's a list of scripture elements associated with the ref? 
___
*[`src/SIL.Machine/Corpora/TextCorpusBase.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZa2Rf0TjHqnfsXK4A:-NuZa2Rf0TjHqnfsXK4B:b-vjexj2) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/TextCorpusBase.cs#L10)):*
> ```Smalltalk
>         public abstract IEnumerable<IText> Texts { get; }
>         public abstract bool IsTokenized { get; }
>         public abstract ScrVers Versification { get; }
> ```

I'm sure you've thought this through better than I can, but is there no more semantically appropriate alternative to having all the corpora have versifications (even if they're just null)? Couldn't you have parallel scripture/non-scripture inheritance structures? Is what you've done preferable because there's no easy way to reuse code across parallel paths like that?
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 182 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZbZYODSlx7M2vOhw4:-NuZbZYODSlx7M2vOhw5:b5gb76j) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/UsfmTextBase.cs#L182)):*
> ```Smalltalk
>                 else if (
>                     (CurrentTextType == ScriptureTextType.Verse && state.IsVerseText)
>                     || CurrentTextType != ScriptureTextType.Verse
> ```

Is there a reason you can't/shouldn't simplify?



_Code snippet:_
```Smalltalk
CurrentTextType != ScriptureTextType.Verse || state.IsVerseText
```
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 218 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZcJQj-jZU_Bqr0YXL:-NuZcJQj-jZU_Bqr0YXM:bsthiew) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/UsfmTextBase.cs#L218)):*
> ```Smalltalk
>                     && (
>                         (CurrentTextType == ScriptureTextType.Verse && state.IsVerseText)
>                         || CurrentTextType != ScriptureTextType.Verse
> ```

Same as note above about simplification.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 42 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZ_L6j5k8h8i9pXu_o:-NuZlyT4384-Fe_jOZCi:bchqxn5) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/ScriptureRef.cs#L42)):*

Why is this called `Path` given it's a list of scripture elements associated with the ref?
</blockquote></details>

I agree. It is not a great name, but I wasn't sure what else to call it. I called it `Path`, because it captures the path to the particular segment in the USFM hierarchy.
___
*[`src/SIL.Machine/Corpora/TextCorpusBase.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZa2Rf0TjHqnfsXK4A:-NuZmIRxAwmpp72Zxwua:bhonfr6) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/TextCorpusBase.cs#L10)):*

I'm sure you've thought this through better than I can, but is there no more semantically appropriate alternative to having all the corpora have versifications (even if they're just null)? Couldn't you have parallel scripture/non-scripture inheritance structures? Is what you've done preferable because there's no easy way to reuse code across parallel paths like that?
</blockquote></details>

I didn't love this change, but I couldn't come up with a better way. I need some way to easily tell if a corpus is a Scripture corpus and uses `ScriptureRef`s. Machine supports operations on corpora, which are implemented through decorator classes. These decorator classes are generic and can be applied to any type of corpus. This obscures the original type of the corpus, so we can't reliably use `is ScriptureTextCorpus` or anything like that. The other way to determine if it is a Scripture corpus was to actually iterate through the rows, grab the ref of the first row, and check if it is a `ScriptureRef`, but that is really hacky. I think this is the best option.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 182 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZbZYODSlx7M2vOhw4:-NuZnpbM5bjcujq5cerr:b-d66m2b) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/UsfmTextBase.cs#L182)):*

Is there a reason you can't/shouldn't simplify?
</blockquote></details>

There isn't really any reason I can't simplify to the suggested conditional. I was just trying to make the intent clearer. I will go ahead and simplify it.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: 33 of 34 files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 182 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZbZYODSlx7M2vOhw4:-NuZpE-O548zpvK4Z8km:bb6un1j) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/UsfmTextBase.cs#L182)):*

There isn't really any reason I can't simplify to the suggested conditional. I was just trying to make the intent clearer. I will go ahead and simplify it.
</blockquote></details>

Done
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 218 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZcJQj-jZU_Bqr0YXL:-NuZpF5f8pc1k9KyOlra:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/UsfmTextBase.cs#L218)):*

Same as note above about simplification.
</blockquote></details>

Done.

Enkidu93 (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 42 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZ_L6j5k8h8i9pXu_o:-NuZyFGxBcomrmCFG1gq:b-uukhzr) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/ScriptureRef.cs#L42)):*

I agree. It is not a great name, but I wasn't sure what else to call it. I called it `Path`, because it captures the path to the particular segment in the USFM hierarchy.
</blockquote></details>

I see. OK. Nothing better comes to mind. Maybe John will have a thought.
___
*[`src/SIL.Machine/Corpora/TextCorpusBase.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZa2Rf0TjHqnfsXK4A:-NuZyTNO5IiuCbaTCbgJ:b1j22q8) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/TextCorpusBase.cs#L10)):*

I didn't love this change, but I couldn't come up with a better way. I need some way to easily tell if a corpus is a Scripture corpus and uses `ScriptureRef`s. Machine supports operations on corpora, which are implemented through decorator classes. These decorator classes are generic and can be applied to any type of corpus. This obscures the original type of the corpus, so we can't reliably use `is ScriptureTextCorpus` or anything like that. The other way to determine if it is a Scripture corpus was to actually iterate through the rows, grab the ref of the first row, and check if it is a `ScriptureRef`, but that is really hacky. I think this is the best option.
</blockquote></details>

That makes sense. This might be even more hacky, but is there some way you could make a helper method that uses reflection to see if the corpus has a versification property? I suppose this isn't really practically so different than having the versification where inapplicable set to null, but could it keep `Versification` from floating around in classes where it doesn't belong?

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 29 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nu_tnNOD4CvvysmN9qM:-Nu_tnNOD4CvvysmN9qN:b9ir5lf) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRef.cs#L29)):*
> ```Smalltalk
>                     path.Add(new ScriptureElement(0, elem[0]));
>                 else
>                     path.Add(new ScriptureElement(int.Parse(elem[0], CultureInfo.InvariantCulture), elem[1]));
> ```

This looks fragile.  There should be a spec that this is designed to.  Either it is general, or there is a specific set of labels that should work in a very specific format.  I notice that something like ABC:123 would break it, in that it would be split into 2, and ABC would not be able to parse into a number.

johnml1135:
Notes from the code coverage:
* Overlaps, comparisons and hash codes untested for ScriptureReference and ScriptureElement
* CreateRow in ScriptureText untested
* Ref in UsfmTextUpdater untested

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 29 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nu_tnNOD4CvvysmN9qM:-NucpWWmBdo4Z56WcWSy:b4mxejt) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRef.cs#L29)):*

This looks fragile.  There should be a spec that this is designed to.  Either it is general, or there is a specific set of labels that should work in a very specific format.  I notice that something like ABC:123 would break it, in that it would be split into 2, and ABC would not be able to parse into a number.
</blockquote></details>

We should make a wiki page explaining how the parsing works - or an extended comment at the beginning of this file.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NucyZMo50Y9belDXSZy:-NucyZMo50Y9belDXSZz:b-mx6hk0) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L16)):*
> ```Smalltalk
>     public abstract class ScriptureRefUsfmParserHandlerBase : UsfmParserHandlerBase
>     {
>         private VerseRef _verseRef;
> ```

This should be _curVerseRef - as it can change.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NucyZMo50Y9belDXSZy:-NucynZ1EGehM_8HAeO1:b-vgipwa) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L16)):*

This should be _curVerseRef - as it can change.
</blockquote></details>

And _curElements

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NucyZMo50Y9belDXSZy:-NudD_uODCQKfHPvqP8g:b-5d0u4) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L16)):*

And _curElements
</blockquote></details>

Hmm. I'm looking at it more.  I am just wondering why you use _curTextType, but not _cur for the other ones...

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 205 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudItN68Qpav9xbvE7L:-NudItN68Qpav9xbvE7M:bfhqo90) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L205)):*
> ```Smalltalk
>             ScriptureElement prevElem = _elements.Pop();
>             _elements.Push(new ScriptureElement(prevElem.Position + 1, marker));
>         }
> ```

I think I got it.  The reference would be:
* Verse reference: MAT 3:12-22
* Elements: Section 4, Paragraph 3, Cell 18.
* Elements as they are in here: Section 4, Section 0, Paragraph 3, Paragraph 0, Cell 18 (0's get removed)
* "Parent elements" can contain others - element "0" symbolically 

There should be an explanation of this (at least some examples) - it took me a while to figure out.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 35 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudXGT5CKmrbwH8wJCW:-NudXGT5CKmrbwH8wJCX:b-y16mh9) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L35)):*
> ```Smalltalk
>             {
>                 EndVerseText(state, CreateVerseRefs());
>                 _curTextType.Pop();
> ```

Why is this popped here?  It would not be used anymore.  I would assume either it would be cleared or left alone...

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 69 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudYSVWEuczFo1Qw6Jd:-NudYSVWEuczFo1Qw6Je:b-dpuv9s) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L69)):*
> ```Smalltalk
>                     EndVerseText(state, CreateVerseRefs());
>                     _curTextType.Pop();
>                 }
> ```

So, for duplicated verses, just take the first one and then ignore all duplicates?  Needs documentation...

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 69 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudYSVWEuczFo1Qw6Jd:-NudYdViA2v_GaOp83dy:boktb4t) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L69)):*

So, for duplicated verses, just take the first one and then ignore all duplicates?  Needs documentation...
</blockquote></details>

Why was the duplicate / merge verse logic needed to be added now?  Was it just moved from somewhere else?  It is not specific to non-verse text...

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 71 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudZ8Sy7A6JOSYiHWFo:-NudZ8Sy7A6JOSYiHWFp:b-al9bam) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L71)):*
> ```Smalltalk
>                 }
>                 _duplicateVerse = true;
>                 UpdateVerseRef(state.VerseRef, marker);
> ```

If it's truly a duplicate verse (and we are just throwing it on the ground), why are we also updating the verse reference?  Wouldn't the marker be the same?  And if it was not, we would want the first marker, not the one that came from the duplicated text.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 60 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudaJtcEDagwVGpjq1r:-NudaJtcEDagwVGpjq1s:b-d1tl17) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L60)):*
> ```Smalltalk
>             string marker,
>             string altNumber,
>             string pubNumber
> ```

What happens for alternative verses in descriptive sections?  They would not be handled as verses.. https://ubsicap.github.io/usfm/chapters_verses/index.html#va-va.  Would /va be counted as a verse?  Would then the description be counted as duplicate?

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 102 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nudbt-b6mCvsBYeMvj9:-Nudbt-b6mCvsBYeMvjA:bgqjsu9) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L102)):*
> ```Smalltalk
>         {
>             if (_verseRef.IsDefault)
>                 UpdateVerseRef(state.VerseRef, marker);
> ```

What is intended here?  Only at the beginning of books do we update from the parser state, but not later?  Why not later?

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 18 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue1EL25UkXOidrgi3m:-Nue1EL25UkXOidrgi3n:b9uklpn) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L18)):*
> ```Smalltalk
>         private VerseRef _verseRef;
>         private readonly Stack<ScriptureElement> _elements;
>         private readonly Stack<ScriptureTextType> _curTextType;
> ```

As another question - _verseRef is never initialized and I assume it starts as a null.  Should this be the case?  Or should it be initialized to default?

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 126 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue7n6f-eH30TAMq6nB:-Nue7n6f-eH30TAMq6nC:b-vec9h7) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L126)):*
> ```Smalltalk
>             if (CurrentTextType == ScriptureTextType.NonVerse)
>                 StartParentElement(marker);
>         }
> ```

So, if it is a verse (or a note), just strip out all formatting?  If there is a paragraph or a table in a verse or cell, squish it together and translate the whole thing?

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 126 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue7n6f-eH30TAMq6nB:-Nue8-C6-Jrru5Wc7d86:b-2bnvqf) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L126)):*

So, if it is a verse (or a note), just strip out all formatting?  If there is a paragraph or a table in a verse or cell, squish it together and translate the whole thing?
</blockquote></details>

Documentation...

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureText.cs` line 104 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueA9HGBv4vLJelzTbF:-NueA9HGBv4vLJelzTbG:bi74y3k) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureText.cs#L104)):*
> ```Smalltalk
>         }
> 
>         protected TextRow CreateRow(
> ```

This appears to never be called

johnml1135:
*[`src/SIL.Machine/Corpora/TextCorpusBase.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZa2Rf0TjHqnfsXK4A:-NueBG8R1Ri89AAql1xn:b-25jh32) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/TextCorpusBase.cs#L10)):*

That makes sense. This might be even more hacky, but is there some way you could make a helper method that uses reflection to see if the corpus has a versification property? I suppose this isn't really practically so different than having the versification where inapplicable set to null, but could it keep `Versification` from floating around in classes where it doesn't belong?
</blockquote></details>

Or, add another property to TextCorpusBase called "CorpusType" or similar that would be hard coded to either a string or an enum of "TextCorpus" or "ScriptureCorpus".  This should survive decorators and be more explicit when trying to determine the underlying type.

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmParser.cs` line 144 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueE9Ch3Iz7wvDbaLpW:-NueE9CiAufSeI-JNSV9:bizaqbv) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmParser.cs#L144)):*
> ```Smalltalk
>             if (State.Index >= State.Tokens.Count - 1)
>             {
>                 CloseAll();
> ```

So, close all USFM markers on the end of USFM (makes sense), the beginnings of notes, chapters and books as well as verses that don't have verse text?  Is that correct?

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 111 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueFKl_BKOzlwy5y4Wb:-NueFKla21VosrVIWAkM:bv3e5qt) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L111)):*
> ```Smalltalk
>                     OutputMarker(state);
>                 }
>                 else if (CurrentTextType == ScriptureTextType.Verse)
> ```

This is an odd dance between the base and parent class - handling non-verse cells in the base and verse cells right here...

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 180 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueIzE20lklw56KeMlM:-NueIzE20lklw56KeMlN:b-pjfatv) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L180)):*
> ```Smalltalk
>                     _rowTexts.Peek().Append("//");
>                 }
>                 else if (CurrentTextType != ScriptureTextType.Verse || state.IsVerseText)
> ```

Ok, so if it isn't a verse or is a verse?  I don't understand...

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 254 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueVYrh1fSfwY7MUAC3:-NueVYrh1fSfwY7MUAC4:b-tx5hk6) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L254)):*
> ```Smalltalk
> 
>                 _rowTexts.Push(new StringBuilder());
>             }
> ```

So, if I understand, If it's a verse, it is all glumbed together.  If it is out of a verse, each part is treated individually.  If markers are included, keep the notes in place.

johnml1135 (commented):
> Reviewed 16 of 34 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: all files reviewed, 21 unresolved discussions (waiting on @ddaspit and @Enkidu93)

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 351 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NuetmcM2yN7swGUGV5L:-NuetmcN8i7-2Kj0RpBk:b-jgqvcz) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L351)):*
> ```Smalltalk
>         private void SkipTokens(UsfmParserState state)
>         {
>             _tokenIndex = state.Index + 1 + state.SpecialTokenCount;
> ```

So, alternate verses and other special tokens will be skipped over if we are replacing the text?  Would we get those from the translated text?  Are we sure we want to blow away the special tokens?

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 268 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nueuda93tiX-M6jAMWL:-NueudaA3QPVkY5oeoGy:b-6mez2s) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L268)):*
> ```Smalltalk
>             {
>                 _tokens.Add(state.Token);
>                 _tokens.Add(new UsfmToken(UsfmTokenType.Character, "ft", null, "ft*"));
> ```

Are we sure there are only footnotes?  What about endnotes?  Don't we have the marker in the Scripture reference that we can pull out and use?

johnml1135 (commented):
> Reviewed 1 of 34 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: all files reviewed, 23 unresolved discussions (waiting on @ddaspit and @Enkidu93)

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: 32 of 34 files reviewed, 22 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 29 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nu_tnNOD4CvvysmN9qM:-NujzrAkExlnC4Ymdv2T:b-g7kdyc) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRef.cs#L29)):*

We should make a wiki page explaining how the parsing works - or an extended comment at the beginning of this file.
</blockquote></details>

Unfortunately, there is no spec for specifying a reference to a particular text segment in a USFM/USX file. I had to make one up that fits our needs. The `ScriptureRef` gets serialized to a specific format. If the format is invalid, then it might fail to parse. 
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 16 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NucyZMo50Y9belDXSZy:-Nuk3opa7T_lc9F6DagQ:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L16)):*

Hmm. I'm looking at it more.  I am just wondering why you use _curTextType, but not _cur for the other ones...
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 18 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue1EL25UkXOidrgi3m:-Nuk-o5ZBFJMOeAf--LR:bb17n9u) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L18)):*

As another question - _verseRef is never initialized and I assume it starts as a null.  Should this be the case?  Or should it be initialized to default?
</blockquote></details>

`VerseRef` is a struct (value type), so it isn't nullable by default.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 35 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudXGT5CKmrbwH8wJCW:-Nuk05xb96S_cxb4CHWn:b-pbfzb) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L35)):*

Why is this popped here?  It would not be used anymore.  I would assume either it would be cleared or left alone...
</blockquote></details>

It doesn't have to be. I was just keeping the logic consistent across the entire class.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 60 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudaJtcEDagwVGpjq1r:-Nuk7wc_0gMW09LYexeC:bkgaj5n) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L60)):*

What happens for alternative verses in descriptive sections?  They would not be handled as verses.. https://ubsicap.github.io/usfm/chapters_verses/index.html#va-va.  Would /va be counted as a verse?  Would then the description be counted as duplicate?
</blockquote></details>

They would be ignored.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 69 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudYSVWEuczFo1Qw6Jd:-Nuk0b5Q9VbPHAfrECvD:bja4wt3) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L69)):*

Why was the duplicate / merge verse logic needed to be added now?  Was it just moved from somewhere else?  It is not specific to non-verse text...
</blockquote></details>

This logic was moved from the `UsfmTextBase` class.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 71 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudZ8Sy7A6JOSYiHWFo:-Nuk41YK8qTa-sc8aKFA:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L71)):*

If it's truly a duplicate verse (and we are just throwing it on the ground), why are we also updating the verse reference?  Wouldn't the marker be the same?  And if it was not, we would want the first marker, not the one that came from the duplicated text.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 102 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nudbt-b6mCvsBYeMvj9:-Nuk44-623WXrZkKDXNa:b-75d3i4) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L102)):*

What is intended here?  Only at the beginning of books do we update from the parser state, but not later?  Why not later?
</blockquote></details>

We only update the verse ref once we hit the first paragraph. Prior to that we are in the book identification section.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 126 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue7n6f-eH30TAMq6nB:-Nuk7LPQEnKtaNxz2_MQ:bdkbqhj) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L126)):*

Documentation...
</blockquote></details>

Yes, that is correct.
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 205 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudItN68Qpav9xbvE7L:-NukASj51nA6-e7jfsnX:b-utaza2) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L205)):*

I think I got it.  The reference would be:
* Verse reference: MAT 3:12-22
* Elements: Section 4, Paragraph 3, Cell 18.
* Elements as they are in here: Section 4, Section 0, Paragraph 3, Paragraph 0, Cell 18 (0's get removed)
* "Parent elements" can contain others - element "0" symbolically 

There should be an explanation of this (at least some examples) - it took me a while to figure out.
</blockquote></details>

I added a description of how scripture refs work in the class documentation. 
___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 230 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NudDk6W2oZ7YIYdE2u7:-Nuk7fKb55cQcoRrR97N:bdkbqhj) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L230)):*

So you create these for Notes (all types), Paragraphs (all types) and Cells (no other types of characters).  Is this the intention (it seems reasonable from little that I see)? This needs to be documented in Swagger.
</blockquote></details>

Yes, that is correct.
___
*[`src/SIL.Machine/Corpora/ScriptureText.cs` line 104 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueA9HGBv4vLJelzTbF:-Nuk55GSDouDnM8r7_0H:bxe5rpw) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureText.cs#L104)):*

This appears to never be called
</blockquote></details>

I left it in as a helper method for classes that extend this abstract base class.
___
*[`src/SIL.Machine/Corpora/TextCorpusBase.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/machine/179#-NuZa2Rf0TjHqnfsXK4A:-Nuk5OaX6fO0bDL0uM43:b-esd5a1) ([raw file](https://github.com/sillsdev/machine/blob/69155866f7ab3e098a10b6b19f88756730b4ecd6/src/SIL.Machine/Corpora/TextCorpusBase.cs#L10)):*

Or, add another property to TextCorpusBase called "CorpusType" or similar that would be hard coded to either a string or an enum of "TextCorpus" or "ScriptureCorpus".  This should survive decorators and be more explicit when trying to determine the underlying type.
</blockquote></details>

@Enkidu93 Reflection should only be used as a last resort.

@johnml1135 Yes, but we would still need to have the `Versification` property, so I figured it was easier to just add this one property and return `null` if it isn't a Scripture corpus. There is an `IsScripture` extension method on `ITextCorpus` that checks this property.
___
*[`src/SIL.Machine/Corpora/UsfmParser.cs` line 144 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueE9Ch3Iz7wvDbaLpW:-Nuk698VBsptExAO470Y:ba7r07s) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmParser.cs#L144)):*

So, close all USFM markers on the end of USFM (makes sense), the beginnings of notes, chapters and books as well as verses that don't have verse text?  Is that correct?
</blockquote></details>

This is a fix for what was essentially a bug in the `UsfmParser` where the `End<element>` methods would not get called at the end of the USFM file. This fix ensures that they get called properly.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 70 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueEilZCzgvDn7050Um:-Nuk8OKkBbuwTseQzHaH:b-nw22fl) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L70)):*

This is where _verseRef was set to default - but not anymore.  But it is still asked if it IsDefault...
</blockquote></details>

I'm not sure I understand what you mean.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 78 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueEsUPBYIRfRrraGuL:-Nuk8RNZ2ai4Zm-C4cMX:b-avrnrg) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L78)):*

This logic is retranscribed - but also duplicates are checked, with odd logic (by my estimation).
</blockquote></details>

I'm not sure what you mean.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 111 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueFKl_BKOzlwy5y4Wb:-Nuk8hRV66dqScxNaxj7:b-rjcpdu) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L111)):*

This is an odd dance between the base and parent class - handling non-verse cells in the base and verse cells right here...
</blockquote></details>

If a table cell occurs inside of a verse, then we need to make sure that we insert a space in the collected verse text.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 180 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueIzE20lklw56KeMlM:-Nuk8uhm64EJBF37RRDq:b-ivtl5f) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L180)):*

Ok, so if it isn't a verse or is a verse?  I don't understand...
</blockquote></details>

The `state.IsVerseText` refers specifically to verse text within a verse. A verse can contain text that isn't verse text.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 254 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueVYrh1fSfwY7MUAC3:-Nuk9NGD43POvOwtK4fB:b-whkzev) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L254)):*

So, if I understand, If it's a verse, it is all glumbed together.  If it is out of a verse, each part is treated individually.  If markers are included, keep the notes in place.
</blockquote></details>

Yes, that sound about right.
___
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 293 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueYXBu1HB4_yteUeUO:-Nuk9Tb43Osga1OyLtfy:b-shbnug) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L293)):*

This logic replaces Sentence completed function.  Has the logic changed?
</blockquote></details>

No, it hasn't. I added this in so that the sentence start property is set correctly for non-verse rows.
___
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 268 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nueuda93tiX-M6jAMWL:-Nuk6mxC49xy2dG5r5_w:b-u89jwg) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L268)):*

Are we sure there are only footnotes?  What about endnotes?  Don't we have the marker in the Scripture reference that we can pull out and use?
</blockquote></details>

This marker is used for both footnotes and endnotes.
___
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 351 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NuetmcM2yN7swGUGV5L:-Nuk9qqcEA_F88fLIbta:b-riqpr7) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L351)):*

So, alternate verses and other special tokens will be skipped over if we are replacing the text?  Would we get those from the translated text?  Are we sure we want to blow away the special tokens?
</blockquote></details>

Yes, if we are skipping a marker, then we should also skip over all of the special tokens that are related to that marker.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 29 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nu_tnNOD4CvvysmN9qM:-NuxcPlg6fbt50rui9wJ:b-w8vwsz) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRef.cs#L29)):*

Unfortunately, there is no spec for specifying a reference to a particular text segment in a USFM/USX file. I had to make one up that fits our needs. The `ScriptureRef` gets serialized to a specific format. If the format is invalid, then it might fail to parse.
</blockquote></details>

I like the description - that tied together with the tests make it understandable for someone coming in for the first time.  We can also add something to Swagger to help people if they pull it from the JSON: https://github.com/sillsdev/serval/issues/356.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 18 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nue1EL25UkXOidrgi3m:-NuxdDxPCuUg-BlBzpFU:b-lt681u) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L18)):*

`VerseRef` is a struct (value type), so it isn't nullable by default.
</blockquote></details>

Ah, the details of Dotnet.

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 70 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueEilZCzgvDn7050Um:-NuxfkUb3_D4Sw77Mvmo:b-tsgk4j) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L70)):*

I'm not sure I understand what you mean.
</blockquote></details>

I misread.

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 78 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-NueEsUPBYIRfRrraGuL:-Nuxgop13WHhYVpPIUV1:bdhoe0r) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextBase.cs#L78)):*

I'm not sure what you mean.
</blockquote></details>

The logic was moved of looking for overlapping verses, etc. from here to ScriptureRefParserHandlerBase.  Also when it was moved duplicate verse checking was modified.  I am ok with it.

johnml1135:
Did you update any tests for the code coverage?

johnml1135 (changes requested):
Updates to test coverage?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179#-:-Nv21XDa92MWJ3WGmG9c:b-25l51p)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 29 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nu_tnNOD4CvvysmN9qM:-Nv8BzEcEA7LvL0wySLk:b9ly05l) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/ScriptureRef.cs#L29)):*

I like the description - that tied together with the tests make it understandable for someone coming in for the first time.  We can also add something to Swagger to help people if they pull it from the JSON: https://github.com/sillsdev/serval/issues/356.
</blockquote></details>

I added a wiki page to explain the parsing: https://github.com/sillsdev/serval/wiki/USFM-Parsing-and-Translation

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 268 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nueuda93tiX-M6jAMWL:-Nv8C_yqEdyqTF32-dOZ:b-mad25l) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L268)):*

We translate everything, we just don't have any way to preserve the markers, so we just put everything in an `ft` marker.
</blockquote></details>

So, here is an example:
```
\f + \fr 7.38: \ft Jesus' words in verses
37-38 may be translated: \fqa “Whoever is thirsty should come to me and drink.
\+fv 38\+fv* As the scripture says, ‘Streams of life-giving water will pour out from
within anyone who believes in me.’”\f*
```
So you are saying that we just translate the whole ft and strip out the \fr, \fqa, \+fv's while keeping the \f and \f* bracketing marks?

johnml1135 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextUpdater.cs` line 268 at r2](https://reviewable.io/reviews/sillsdev/machine/179#-Nueuda93tiX-M6jAMWL:-NvCi6QJBZrpopwfCyrL:b-4c6y0e) ([raw file](https://github.com/sillsdev/machine/blob/4387991240632dd32e213d7becd43d4b3f48a3d5/src/SIL.Machine/Corpora/UsfmTextUpdater.cs#L268)):*

So, here is an example:
```
\f + \fr 7.38: \ft Jesus' words in verses
37-38 may be translated: \fqa “Whoever is thirsty should come to me and drink.
\+fv 38\+fv* As the scripture says, ‘Streams of life-giving water will pour out from
within anyone who believes in me.’”\f*
```
So you are saying that we just translate the whole ft and strip out the \fr, \fqa, \+fv's while keeping the \f and \f* bracketing marks?
</blockquote></details>

Whatever it is is good enough.  If it becomes an issue, we can make a later improvement.

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

johnml1135:
I'll make a new issue for test coverage - https://github.com/sillsdev/machine/issues/181

johnml1135 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/179#-:-NvD_9031jv97Q8Kit3D:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #191, opened 2024-04-22
Usfm parse error logging

johnml1135:
Fixes:
https://github.com/sillsdev/serval/issues/376
https://github.com/sillsdev/serval/issues/375

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/191)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ src/sentencepiece4cbuild/
 !samples/**/release
 **/*.feature.cs
 tests/CoverageResults/*
+tests/SIL.Machine.Tests/Corpora/TestData/usfm/source/*
+tests/SIL.Machine.Tests/Corpora/TestData/usfm/target/*
+
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -83,7 +83,7 @@ namespace SIL.Machine.Corpora
             if (_curVerseRef.IsDefault)
                 UpdateVerseRef(state.VerseRef, marker);
 
-            if (!state.IsVerseText)
+            if (!state.IsVerseText || CurrentTextType == ScriptureTextType.NonVerse)
             {
                 StartParentElement(marker);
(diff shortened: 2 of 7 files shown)


ddaspit (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/191)* status: all files reviewed, 1 unresolved discussion (waiting on @johnml1135)

___
*[`src/SIL.Machine/Corpora/UsfmFileText.cs` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine/191#-Nw5tXsn8-9YnWClPRQA:-Nw5tXsn8-9YnWClPRQB:b-7aq3ye) ([raw file](https://github.com/sillsdev/machine/blob/55f36ddc1de62f7d8173988931b541acc826ada6/src/SIL.Machine/Corpora/UsfmFileText.cs#L48)):*
> ```Smalltalk
>                 }
>             }
>             // It's not in the file contents, let's just pull it from the filename
> ```

I don't believe that this will fix the issue. The crash is occurring in `UsfmZipText`. Also, this is not generic enough, since the file name in a PT project is configurable. We should use the file naming setting in the PT project settings to determine how to properly parse the file name. I think we should wait until we get more information before we submit a fix for this issue.

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmFileText.cs` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine/191#-Nw5tXsn8-9YnWClPRQA:-Nw64sc603YtKk6gSn6x:b33t68r) ([raw file](https://github.com/sillsdev/machine/blob/55f36ddc1de62f7d8173988931b541acc826ada6/src/SIL.Machine/Corpora/UsfmFileText.cs#L48)):*

I don't believe that this will fix the issue. The crash is occurring in `UsfmZipText`. Also, this is not generic enough, since the file name in a PT project is configurable. We should use the file naming setting in the PT project settings to determine how to properly parse the file name. I think we should wait until we get more information before we submit a fix for this issue.
</blockquote></details>

Fine enough - I reverted the fix.  The other work is more important for today.

PR #205, opened 2024-05-29
Throw error when both languages are unknown and there's no data

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/284

A couple questions regarding the solution:
1. I noticed that John has it throwing an `OperationCanceledException` when the engine doesn't exist in `NmtPreprocessBuildJob`. Is that what we want? Or should this throw some other exception that would cause the build to be marked as failed instead? And if so, should it also do the same in this case? Maybe I'm missing something. 
2. I thought I'd add a test to cover the logic mentioned above, but it seems to be unreachable because if the engine does not exist, it'll get cut short here in the `HangfireBuildJob` code:  
                `if (!await BuildJobService.BuildJobStartedAsync(engineId, buildId, cancellationToken))
                {
                    completionStatus = JobCompletionStatus.Canceled;
                    return;
                }`
What am I missing?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/205)

### DIFF
--- a/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs
+++ b/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs
@@ -80,7 +80,7 @@ public class NmtPreprocessBuildJob : HangfireBuildJob<IReadOnlyList<Corpus>>
         if (trainCount == 0 && !tagsInFlores200)
         {
             throw new OperationCanceledException(
-                $"Neither language code in build {buildId} are known to the base model, and the data specified for training was empty. Build Canceled."
+                $"Neither language code in build {buildId} are known to the base model, and the data specified for training was empty. Build canceled."
             );
         }
 
--- a/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs
+++ b/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs
@@ -71,11 +71,18 @@ public class NmtPreprocessBuildJob : HangfireBuildJob<IReadOnlyList<Corpus>>
         if (engine is null)
             throw new OperationCanceledException($"Engine {engineId} does not exist.  Build canceled.");
 
-        _languageTagService.ConvertToFlores200Code(engine.SourceLanguage, out string srcLang);
+        bool tagsInFlores200 = _languageTagService.ConvertToFlores200Code(engine.SourceLanguage, out string srcLang);
         buildPreprocessSummary.Add("SourceLanguageResolved", srcLang);
-        _languageTagService.ConvertToFlores200Code(engine.TargetLanguage, out string trgLang);
+        tagsInFlores200 &= _languageTagService.ConvertToFlores200Code(engine.TargetLanguage, out string trgLang);
         buildPreprocessSummary.Add("TargetLanguageResolved", trgLang);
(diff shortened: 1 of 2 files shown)


ddaspit (changes requested):
1. The only reason why an engine wouldn't exist during a build is if the engine was deleted. Deleting an engine while a build is in progress means that the build should be canceled. The case you are covering here is different, so I wouldn't throw an `OperationCanceledException`. `InvalidOperationException` would be appropriate.
2. An engine could be deleted at any point while a build is running. Whenever we get the engine model from the DB, we also check to make sure it still exists as a sanity check.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/205#-:-Nz51i6G89vL_DFbePFw:bg3c3bq)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 76 at r1](https://reviewable.io/reviews/sillsdev/machine/205#-Nz52hLdDug_PQWkarN5:-Nz52hLdDug_PQWkarN6:b39dygn) ([raw file](https://github.com/sillsdev/machine/blob/d13d83fa5d6f536fa1bc59e41bd56c2d103da260/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L76)):*
> ```Smalltalk
>         bool tagsInFlores200 = _languageTagService.ConvertToFlores200Code(engine.SourceLanguage, out string srcLang);
>         buildPreprocessSummary.Add("SourceLanguageResolved", srcLang);
>         tagsInFlores200 &= _languageTagService.ConvertToFlores200Code(engine.TargetLanguage, out string trgLang);
> ```

Instead of using a single boolean, I would separate these into two separate boolean variables and check them both in the `if` statement, I think that would be more clear.

Enkidu93 (commented):
That makes a lot of sense. Thank you! I figured I was missing something. Went ahead and fixed the exception.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/205#-:-Nz5Iehj1zjir7PA0LvW:b-gef5jf)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs` line 76 at r1](https://reviewable.io/reviews/sillsdev/machine/205#-Nz52hLdDug_PQWkarN5:-Nz5Id_PEz3uKGXCN2Z3:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d13d83fa5d6f536fa1bc59e41bd56c2d103da260/src/SIL.Machine.AspNetCore/Services/NmtPreprocessBuildJob.cs#L76)):*

Instead of using a single boolean, I would separate these into two separate boolean variables and check them both in the `if` statement, I think that would be more clear.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/205#-:-NzAEZKA7XnFrMmjTE_E:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #208, opened 2024-06-07
Fix terms bugs

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/391
Fixes https://github.com/sillsdev/serval/issues/392

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/208)


Enkidu93:
I'm not sure of the cleanest way to extend our tests for the case of the nonexistent ProjectBiblicalTerms file. Should I not worry about it or go ahead and create an entire second project to cover the new case?

PR #213, opened 2024-06-18
auto-add missing verses.

johnml1135:
Auto add missing verses -  https://github.com/sillsdev/serval/issues/414

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/213)

### DIFF
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -15,6 +15,10 @@ namespace SIL.Machine.Corpora
     public abstract class ScriptureRefUsfmParserHandlerBase : UsfmParserHandlerBase
     {
         private VerseRef _curVerseRef;
+        protected VerseRef CurVerseRef
+        {
+            get { return _curVerseRef; }
+        }
         private readonly Stack<ScriptureElement> _curElements;
         private readonly Stack<ScriptureTextType> _curTextType;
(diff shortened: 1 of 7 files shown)


ddaspit (changes requested):
I believe that the core underlying issue is that the `ScriptureRefUsfmParserHandlerBase` class ignores duplicate verses. Maybe we should reevaluate this behavior, since the assumption was that this was a mistake in the USFM. The examples from this issue were intentional. If we change the underlying behavior, it should fix this issue.

> Reviewed 4 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/213#-:-O-llRTU9X5ZQvJhlC_R:b-j3fn3m)* status: 4 of 9 files reviewed, all discussions resolved

johnml1135:
I don't fully understand your rationale.  I updated the code to reflect the pattern of the other PR.  It appears to respond intelligently to the test cases added.

ddaspit:
https://github.com/sillsdev/serval/issues/414 is not something that Scripture Forge has asked for, so I think we should defer it for now. In the case of https://github.com/sillsdev/serval/issues/411, I would say that Serval is actually working as intended. We should wait for feedback on how Scripture Forge wants to handle that issue for continuing.

johnml1135:
I pulled out the 2 fixes into 2 different pull requests.  Now this is just add missing verses.  The other issue is handled elsewhere.

johnml1135:
This is no longer needed and being supplanted by giving the user the choice to use the source USFM structure.

PR #217, opened 2024-06-26
Allow USFM texts to be generated from the USFM string only

johnml1135:
https://github.com/sillsdev/serval/issues/415

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/217)

### DIFF
--- a/src/SIL.Machine/Corpora/MemoryStreamContainer.cs
+++ b/src/SIL.Machine/Corpora/MemoryStreamContainer.cs
@@ -7,15 +7,17 @@ namespace SIL.Machine.Corpora
     public class MemoryStreamContainer : DisposableBase, IStreamContainer
     {
         private readonly string _usfm;
+        private readonly Encoding _encoding;
 
-        public MemoryStreamContainer(string usfm)
+        public MemoryStreamContainer(string usfm, Encoding encoding)
         {
             _usfm = usfm;
+            _encoding = encoding;
         }
 
         public Stream OpenStream()
         {
-            return new MemoryStream(Encoding.UTF8.GetBytes(_usfm));
+            return new MemoryStream(_encoding.GetBytes(_usfm));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


ddaspit (commented):
What is this for?

> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/217#-:-O0L16nl0iWo_EUyLgmM:bzbe37d)* status: 1 of 4 files reviewed, all discussions resolved

johnml1135:
For the new source USFM changes in serval.

ddaspit:
I don't quite understand why this is necessary. All we have to do is force it to go down the use source book code path in [PretranslationService.GetUsfmAsync()](https://github.com/sillsdev/serval/blob/f425269d45330c04d3e5ebed6787fcf6ee1c1f1c/src/Serval/src/Serval.Translation/Services/PretranslationService.cs#L121).

johnml1135:
To import the target text into the source USFM, they need to have a `ParallelTextCorpus` to create the `targetTextAsPretranslations`.  That `ParallelTextCorpus` needs to be created from only the USFM text and the settings (instead of the whole zip files - we are wanting this to be lightweight) and so, we need the ability to create a corpus from merely a few USFM strings.  The other code is just for handling the UTF encoding a bit more cleanly (to my understanding), due to using strings rather than bytes.

ddaspit:
We don't need to transfer the target text into the source USFM. The SF team said that wasn't a requirement.

johnml1135:
Yet we said that we would add the option "use Source USFM" instead of just adding another choice.  If we will use he source USFM structure but have "only existing", how will we do that?  I believe the solution built with full tests should work fine and I recommend just keeping as is.

ddaspit (commented):
It unnecessarily adds complexity to the implementation, which increases the maintenance cost for no benefit. The existing implementation won't change, so what "only existing" means is the same as it was before. I'm not sure I see the problem.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/217#-:-O0jXvXtDzYmNFtW6-S2:b3p11sw)* status: 1 of 4 files reviewed, all discussions resolved

ddaspit:
I am closing this PR, since we don't need this change.

PR #233, opened 2024-08-15
Use terms localizations

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/240

Parallel PR: https://github.com/sillsdev/serval/pull/454

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)


ddaspit (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine/233#-O4MZtvU3mW8ViHNqjQ2:-O4MZtvU3mW8ViHNqjQ3:b-fgnrh) ([raw file](https://github.com/sillsdev/machine/blob/056c9fd209b67ae0979be412915b3b4078f0a9db/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L48)):*
> ```Smalltalk
>                 bool useTermsRenderingXml = !preferTermsLocalization && termsFileEntry != null;
> 
>                 if (!SupportedLanguageTermsLocalizationXmls.TryGetValue(languageCode, out string resourceName))
> ```

If `languageCode` is `null`, we should use the language code specified in the PT settings file.

Also, the localizations only apply to the Major Biblical Terms list.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 171 at r1](https://reviewable.io/reviews/sillsdev/machine/233#-O4MaRmBAHDV5zqeWyoO:-O4MaRmBAHDV5zqeWyoP:by42f6a) ([raw file](https://github.com/sillsdev/machine/blob/056c9fd209b67ae0979be412915b3b4078f0a9db/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L171)):*
> ```Smalltalk
>         {
>             //If entire term rendering is surrounded in square brackets, remove them
>             Regex rx = new Regex(@"^\[(.+?)\]$", RegexOptions.Compiled);
> ```

We should take the opportunity to make this regex a constant.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: 8 of 13 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 48 at r1](https://reviewable.io/reviews/sillsdev/machine/233#-O4MZtvU3mW8ViHNqjQ2:-O4QTaprCVGTvIn5xVGB:b-3vjk0f) ([raw file](https://github.com/sillsdev/machine/blob/056c9fd209b67ae0979be412915b3b4078f0a9db/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L48)):*

If `languageCode` is `null`, we should use the language code specified in the PT settings file.

Also, the localizations only apply to the Major Biblical Terms list.
</blockquote></details>

I've gone ahead and just used the settings language code. Let me know if you think I should also provide a route to specify a code. It seemed to me that one or the other would simplify things and be cleaner.

Thank you, yes! Done.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 171 at r1](https://reviewable.io/reviews/sillsdev/machine/233#-O4MaRmBAHDV5zqeWyoO:-O4QTsnGCv9hR3SfUN9B:bhu7mhk) ([raw file](https://github.com/sillsdev/machine/blob/056c9fd209b67ae0979be412915b3b4078f0a9db/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L171)):*

We should take the opportunity to make this regex a constant.
</blockquote></details>

Done. Also, with the other. Also, uncommented a line that I accidentally left commented..

Enkidu93 (commented):
I'm going to rework the testing a little bit since it's awkward to test everything I need to. I'll ping you know when I'm ready for more review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4QV54c6YVs-FCDn0Hi:b-pvi8og)* status: 8 of 13 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

Enkidu93 (commented):
@ddaspit, here's one solution. Let me know what you think. There are other options to make testing easier, but this seemed like the cleanest. I changed the name for the terms corpus to make it more consistent with the newer classes operating on Paratext projects. Do you think that's appropriate? If so, should I change the text corpus as well? A similar strategy could be used there to make testing more flexible. If that does happen, there might also be a way to change the inheritance/implementing to make it overall cleaner across classes that interact with Paratext projects.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4RrFrt4eYP4fGm7f8x:b-fmo5cj)* status: 8 of 16 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @johnml1135)

ddaspit (commented):
> Reviewed 3 of 5 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/MemoryParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4S-OgH6nWtdZ0lEq5C:-O4S-OgH6nWtdZ0lEq5D:bau7rc8) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/MemoryParatextProjectTermsCorpus.cs#L7)):*
> ```Smalltalk
> {
>     public class MemoryParatextProjectTermsCorpus : ParatextProjectTermsCorpusBase
> ```

This feels like a test mock and not a practically useful class. I would move it to the test library.
___
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4Rvx244R43rR7Krbu0:b-bdgbx4) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*
> ```Smalltalk
> {
>     public class ZipParatextProjectTermsCorpus : ParatextProjectTermsCorpusBase
> ```

I would prefer to keep the same name with the same constructor parameters so that it isn't a breaking change. The base class can be renamed to `ParatextTermsCorpusBase` to stay consistent.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 37 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4S05q35LEcz3uT2xO6:-O4S05q35LEcz3uT2xO7:b-kyiii8) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L37)):*
> ```Smalltalk
>         private static readonly Regex NumericalInformationRegex = new Regex(@"\s+\d+(\.\d+)*$", RegexOptions.Compiled);
> 
>         public ParatextProjectTermsCorpusBase(ParatextProjectSettings settings)
> ```

Constructors for abstract base classes should be `protected`.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 42 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rzh-S62_Ggr7CuC5H:-O4Rzh-S62_Ggr7CuC5I:b-lsj6pb) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L42)):*
> ```Smalltalk
>         }
> 
>         protected void AddTexts(IEnumerable<string> termCategories, bool preferTermsLocalization = false)
> ```

It would be cleaner to move the logic in this method to the constructor.

I would also change the behavior, so that it uses the rendering if it exists for a term and falls back to the gloss if it doesn't. I would rename `preferTermsLocalization` to `fallbackToTermGloss` to match the behavior change.

Enkidu93 (commented):
@ddaspit , I'm assuming from your lack of general comments that you do think this an appropriate way to enable better testing; is that correct?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4SGdGO5ZmgO5Hah0aS:b-s18bde)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4SGPje-elq4oXP-6pL:bk6taso) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*

I would prefer to keep the same name with the same constructor parameters so that it isn't a breaking change. The base class can be renamed to `ParatextTermsCorpusBase` to stay consistent.
</blockquote></details>

OK, what about the `ZipParatext...` classes? Should we perhaps use the 'Backup' naming scheme there too?
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 42 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rzh-S62_Ggr7CuC5H:-O4SFxQGCxZwHifnlscN:b-ozg0d4) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L42)):*

It would be cleaner to move the logic in this method to the constructor.

I would also change the behavior, so that it uses the rendering if it exists for a term and falls back to the gloss if it doesn't. I would rename `preferTermsLocalization` to `fallbackToTermGloss` to match the behavior change.
</blockquote></details>

It was previously in the constructor. The issue is that in the zip child class, the `_archive` field needs to be set before running the constructor logic. Thoughts?

Enkidu93:
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 42 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rzh-S62_Ggr7CuC5H:-O4evyzE5I0_6WwlLwrP:bl5kitg) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L42)):*

It was previously in the constructor. The issue is that in the zip child class, the `_archive` field needs to be set before running the constructor logic. Thoughts?
</blockquote></details>

Also, regarding the functionality change, I get what you're saying (move things to the term-level basically) but I don't understand the name. Wouldn't you always want to fall back to the gloss if there isn't a rendering and there is a gloss? If not, then would it make more sense to control it from a variable with a name more like `useTermsLocalizations`?

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4fvmsACOQYjSuLfqsc:bwy2814) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*
> ```Smalltalk
>                 );
>             }
>             string languageCode = languageIsoCodeSettingParts[0];
> ```

So we want to crash if there is no language code?  It makes general sense - but I wonder about all paratext projects.  Can we test a bunch to make sure that we are not excluding some DBL resources that don't have the code set?  Can we set it to a meaningless value instead?  Should we?

ddaspit (commented):
Yes, your approach for testing is sound.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4ftM9m4_ZD6ryojzqt:b-ceqalv)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4fuLOm5G0HSP-hkRea:bmfflr0) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*

OK, what about the `ZipParatext...` classes? Should we perhaps use the 'Backup' naming scheme there too?
</blockquote></details>

Yes, this class should be called `ParatextBackupTermsCorpus`, so that it has the same name as before.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 42 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rzh-S62_Ggr7CuC5H:-O4ftaT55OUONE24QyOp:b101a9p) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L42)):*

Also, regarding the functionality change, I get what you're saying (move things to the term-level basically) but I don't understand the name. Wouldn't you always want to fall back to the gloss if there isn't a rendering and there is a gloss? If not, then would it make more sense to control it from a variable with a name more like `useTermsLocalizations`?
</blockquote></details>

Of course, I see the issue now. I would just get rid of the constructor altogether and pass the `settings` directly into the `AddTexts` method. 

I do think there are cases where you wouldn't want to use the term glosses. The name is fine. I would just refer to the "localizations" as "glosses", i.e. `useTermGlosses`.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4fwR8K1HJTrNJVPPOi:bs7rtw1) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

So we want to crash if there is no language code?  It makes general sense - but I wonder about all paratext projects.  Can we test a bunch to make sure that we are not excluding some DBL resources that don't have the code set?  Can we set it to a meaningless value instead?  Should we?
</blockquote></details>

A Paratext project should have a defined language code setting. Of course, we have learned that anything can happen with a Paratext project, so I don't have a problem with setting the language code to `null` if it isn't defined. We will need to check for null when we use it.

johnml1135 (commented):
> Reviewed 8 of 10 files at r1, 3 of 5 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93)

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4fxdmG4sxc3ATZF6UX:b-sg1qn4) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

A Paratext project should have a defined language code setting. Of course, we have learned that anything can happen with a Paratext project, so I don't have a problem with setting the language code to `null` if it isn't defined. We will need to check for null when we use it.
</blockquote></details>

And add a test to make sure it is handled gracefully.  I believe it would be better to continue on than to crash in this scenario.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4fz-UM7iJWQZDjDTpo:b-11hobi) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

And add a test to make sure it is handled gracefully.  I believe it would be better to continue on than to crash in this scenario.
</blockquote></details>

Alright.
___
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4fxOMD9Qb-mbkG-k-z:bi7n40m) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*

Yes, this class should be called `ParatextBackupTermsCorpus`, so that it has the same name as before.
</blockquote></details>

I understand that. I'm asking about the inconsistency of this name with `ZipParatextProjectTextUpdater`, `ZipParatextProjectSettingsParser`, and all the related classes. The reason I changed the name was to make it more consistent with them. Happy to change it back, but I'm wondering if we should also change the above classes to `ParatextBackupTextUpdater` etc. We kinda have two naming schemes going.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 42 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rzh-S62_Ggr7CuC5H:-O4fyp8k9lc8NCxF1MdC:b2nhktp) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L42)):*

Of course, I see the issue now. I would just get rid of the constructor altogether and pass the `settings` directly into the `AddTexts` method. 

I do think there are cases where you wouldn't want to use the term glosses. The name is fine. I would just refer to the "localizations" as "glosses", i.e. `useTermGlosses`.
</blockquote></details>

OK.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 5 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4g1QMO-iobPos2tA5p:b1l2gek) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*

I understand that. I'm asking about the inconsistency between `ZipParatextProjectTextUpdater`, `ZipParatextProjectSettingsParser`, and all the related classes. The reason I changed the name was to make it more consistent with them. Happy to change it back, but I'm wondering if we should also change the above classes to `ParatextBackupTextUpdater` etc. We kinda have two naming schemes going.
</blockquote></details>

You can leave the names the way they are. It is unfortunate that they follow different naming schemes, but I do think of the corpus classes and the parser/updater classes differently. The corpus classes are higher-level classes that a developer/researcher would consume when creating an NLP pipeline, so their names are intended to clearly communicate the type of corpus. For example, the name `ParatextBackupTextCorpus` indicates to the user that this is the class to use when he wants to read a Paratext project backup file. The parser/updater classes are lower-level building blocks that are used to implement the higher-level classes, so their names reflect that.

Enkidu93:
*[`src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4Rvx244R43rR7Krbu-:-O4gMonzEoDMz1JwMacn:b-m27znj) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ZipParatextProjectTermsCorpus.cs#L7)):*

You can leave the names the way they are. It is unfortunate that they follow different naming schemes, but I do think of the corpus classes and the parser/updater classes differently. The corpus classes are higher-level classes that a developer/researcher would consume when creating an NLP pipeline, so their names are intended to clearly communicate the type of corpus. For example, the name `ParatextBackupTextCorpus` indicates to the user that this is the class to use when he wants to read a Paratext project backup file. The parser/updater classes are lower-level building blocks that are used to implement the higher-level classes, so their names reflect that.
</blockquote></details>

OK, that's fair. `Backup` is more user-friendly maybe. I do wonder what harm there is in changing the `Zip...` classes to use the `Backup` naming scheme, but it's certainly not pressing. There is some marked similarity at this point though now between the structure of those classes and these.

johnml1135:
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4jjQeR-GgIZZC4FX7m:b9qyb8y) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

Alright.
</blockquote></details>

@ddaspit  - you understand the range of Paratext projects better.  What do you think the default behavior should be?  We have no back communication (right now) other than crashing.  We could also add a step in our onboarding procedure to make sure that the keyterms are set.  We could also add an endpoint to do a "pre-build check" that would verify all of these things.

ddaspit (commented):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4l1v529Y8sscRVAi-Q:b-hynewj) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

@ddaspit  - you understand the range of Paratext projects better.  What do you think the default behavior should be?  We have no back communication (right now) other than crashing.  We could also add a step in our onboarding procedure to make sure that the keyterms are set.  We could also add an endpoint to do a "pre-build check" that would verify all of these things.
</blockquote></details>

Right now, we are only using it for term glosses. So, if the language code isn't set, we don't use glosses, which is what Eli implemented.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l2hy06oOdYorzLQVu:-O4l2hy06oOdYorzLQVv:b-caff6i) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L12)):*
> ```Smalltalk
> 
>         public ParatextBackupTermsCorpus(
>             ZipArchive archive,
> ```

The constructor should accept the path to the zip file.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 14 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l2s7E2LoWRK9qWGKq:-O4l2s7E2LoWRK9qWGKr:bn80f84) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L14)):*
> ```Smalltalk
>             ZipArchive archive,
>             IEnumerable<string> termCategories,
>             bool preferTermsLocalization = false
> ```

This parameter should be called `useTermGlosses`.
___
*[`src/SIL.Machine/Corpora/ParatextTermsCorpusBase.cs` line 169 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l4ULz2eBv1Y5Y-WLD:-O4l4ULz2eBv1Y5Y-WLE:bgwt2lg) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextTermsCorpusBase.cs#L169)):*
> ```Smalltalk
>             //Prefer renderings to gloss localizations
>             IDictionary<string, IReadOnlyList<string>> glosses = termsRenderings
>                 .Concat(termsGlosses.Where(kvp => !termsGlosses.ContainsKey(kvp.Key)))
> ```

I don't think the `Where` predicate is correct. Shouldn't it be `!termRenderings.ContainsKey(kvp.Key)`.

Enkidu93 (commented):
More fixes to come. Just wanted to commit what I had so far. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4l2kuR5RqzxxxovysN:b-f6jfsc)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs` line 113 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4fvmsACOQYjSuLfqsb:-O4kkzJc-okIIFCQWsNc:b-w4mn5k) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectSettingsParserBase.cs#L113)):*

@ddaspit  - you understand the range of Paratext projects better.  What do you think the default behavior should be?  We have no back communication (right now) other than crashing.  We could also add a step in our onboarding procedure to make sure that the keyterms are set.  We could also add an endpoint to do a "pre-build check" that would verify all of these things.
</blockquote></details>

(For now, I'll allow it not to be set, barring a new development, but yes, if we do allow such a thing, it might be helpful to have some artifact that things weren't set - even just logging something)
___
*[`src/SIL.Machine/Corpora/MemoryParatextProjectTermsCorpus.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4S-OgH6nWtdZ0lEq5C:-O4kkswHBpCjOmntqKtu:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/MemoryParatextProjectTermsCorpus.cs#L7)):*

This feels like a test mock and not a practically useful class. I would move it to the test library.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs` line 37 at r3](https://reviewable.io/reviews/sillsdev/machine/233#-O4S05q35LEcz3uT2xO6:-O4klI-Y106m8MtlS6gc:b-7mjkwe) ([raw file](https://github.com/sillsdev/machine/blob/ec31107bd421cbcb187bf68153c1329ab8a3a744/src/SIL.Machine/Corpora/ParatextProjectTermsCorpusBase.cs#L37)):*

Constructors for abstract base classes should be `protected`.
</blockquote></details>

Done - in a sense haha

Enkidu93 (commented):
Sorry, @ddaspit , that was a mid fix commit like I said above. I think I addressed most of your review comments besides the archive vs directory name issue. Let me know.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4lA-iU9kLvIJu_rp2L:bn6gigu)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

Enkidu93 (commented):
(Double sorry - apparently the message to wait on reviewing didn't go through earlier)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4lAIgWEeQyRK3R9Hwd:bp76ew9)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l2hy06oOdYorzLQVu:-O4lBKLg4bZqRLMaIWKj:b-glmamr) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L12)):*

The constructor should accept the path to the zip file.
</blockquote></details>

What's the safest/most appropriate way to handle this? Make this class disposable or just open, close, and reopen the archive as needed when `Exist` and `Open` are called?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @Enkidu93 and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l2hy06oOdYorzLQVu:-O4lD5BR0y-Anfv4Fvzr:b-k98q83) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L12)):*

What's the safest/most appropriate way to handle this? Make this class disposable or just open, close, and reopen the archive as needed when `Exist` and `Open` are called?
</blockquote></details>

Hmm, that is an issue. I would prefer that the class isn't disposable. You could refactor `ParatextTermsCorpusBase` into `ParatextTermsParserBase`. This would be patterned after the `ParatextProjectSettingsParserBase`. It would provide a `Parse` method that returns `IEnumerable<(string, string)> You would then implement a `ZipParatextTermsParser` class that is used by `ParatextBackupTermsCorpus`.

johnml1135 (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: 12 of 16 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Enkidu93)

ddaspit (commented):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSk2Y36rfMpH0DUlV:-O4qSk2Y36rfMpH0DUlW:bnowsq9) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L17)):*
> ```Smalltalk
>             using (var archive = ZipFile.OpenRead(fileName))
>             {
>                 ParatextProjectSettings settings = new ZipParatextProjectSettingsParser(archive).Parse();
> ```

I would just perform the settings parsing in the constructor of `ZipParatextTermsParser`. See `ParatextProjectTextUpdaterBase` for an example.
___
*[`src/SIL.Machine/Corpora/ParatextTermsParserBase.cs` line 37 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qRA0X1NCKtGnMJIJ4:-O4qRA0YCT0Dis-fRPND:b-aufqjs) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextTermsParserBase.cs#L37)):*
> ```Smalltalk
>         private static readonly Regex NumericalInformationRegex = new Regex(@"\s+\d+(\.\d+)*$", RegexOptions.Compiled);
> 
>         public IEnumerable<(string, IEnumerable<string>)> Parse(
> ```

I would make the return type `IEnumerable<(string TermId, IReadOnlyList<string> Glosses)>`.
___
*[`tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs` line 8 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSFo_9YzhFzLTNu8D:-O4qSFo_9YzhFzLTNu8E:b5lskdd) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs#L8)):*
> ```Smalltalk
> 
> [TestFixture]
> public class ParatextTermsCorpusTests
> ```

Really what you are testing here is `ParatextTermsParserBase`. It would be better if you tested the `MemoryParatextTermsParser` class directly and just got rid of the `ParatextProjectTermsCorpus` class entirely.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 12 at r4](https://reviewable.io/reviews/sillsdev/machine/233#-O4l2hy06oOdYorzLQVu:-O4qVmmu1qCHoWCa1smt:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/575641f8f424b68c7238ec4dbbf0c7424844eabd/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L12)):*

Hmm, that is an issue. I would prefer that the class isn't disposable. You could refactor `ParatextTermsCorpusBase` into `ParatextTermsParserBase`. This would be patterned after the `ParatextProjectSettingsParserBase`. It would provide a `Parse` method that returns `IEnumerable<(string, string)> You would then implement a `ZipParatextTermsParser` class that is used by `ParatextBackupTermsCorpus`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSk2Y36rfMpH0DUlV:-O4qWKfQ8Vc6pCRYSfga:baw0wq2) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L17)):*

I would just perform the settings parsing in the constructor of `ZipParatextTermsParser`. See `ParatextProjectTextUpdaterBase` for an example.
</blockquote></details>

But the settings is needed here to create the text id. Related to my other comment, could we just return the text rows themselves? That would solve this as well.
___
*[`tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs` line 8 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSFo_9YzhFzLTNu8D:-O4qUEupCBQiD4yfgNvH:b-al11kz) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs#L8)):*

Really what you are testing here is `ParatextTermsParserBase`. It would be better if you tested the `MemoryParatextTermsParser` class directly and just got rid of the `ParatextProjectTermsCorpus` class entirely.
</blockquote></details>

That's fair. It means that those few lines in the `ParatextBackupTermsCorpus` won't be tested. Is there a reason I couldn't just return the `TextRows` instead of the list of ids and glosses?

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSk2Y36rfMpH0DUlV:-O4qXLt21X1v6Nzy3OAt:blfslvg) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L17)):*

But the settings is needed here to create the text id. Related to my other comment, could we just return the text rows themselves? That would solve this as well.
</blockquote></details>

Good point. I missed that. You can provide the option to pass the settings to the constructor of `ZipParatextTermsParser`. If the settings is null, then you can use the settings parser, similar to what we have in `ParatextProjectTextUpdaterBase`.
___
*[`src/SIL.Machine/Corpora/ParatextTermsParserBase.cs` line 12 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4q_PUy56VumIi5TrDx:-O4q_PUy56VumIi5TrDy:bpa1y1l) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextTermsParserBase.cs#L12)):*
> ```Smalltalk
> {
>     public abstract class ParatextTermsParserBase
> ```

You should rename this to `ParatextProjectTermsParserBase` to match the other PT project classes.
___
*[`tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs` line 8 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSFo_9YzhFzLTNu8D:-O4qYonU4oCHpBNFIagS:bn1zi8j) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs#L8)):*

That's fair. It means that those few lines in the `ParatextBackupTermsCorpus` won't be tested. Is there a reason I couldn't just return the `TextRows` instead of the list of ids and glosses?
</blockquote></details>

This test fixture isn't testing the `ParatextBackupTermsCorpus` class. It is testing the mock. If you want to test the constructor of the `ParatextBackupTermsCorpus` class, then create a separate test fixture for that class. You should be able to use one of the existing test projects for those tests.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit)

___
*[`tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs` line 8 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSFo_9YzhFzLTNu8D:-O4qjWfTAh8iqnHysaNR:bzh2o34) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/tests/SIL.Machine.Tests/Corpora/ParatextTermsCorpusTests.cs#L8)):*

This test fixture isn't testing the `ParatextBackupTermsCorpus` class. It is testing the mock. If you want to test the constructor of the `ParatextBackupTermsCorpus` class, then create a separate test fixture for that class. You should be able to use one of the existing test projects for those tests.
</blockquote></details>

True, I know it's not really testing it. I was just exploring the idea of returning the rows themselves to consolidate the logic and also simplify the issue above, but I can just do this.

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233)* status: 12 of 19 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @johnml1135)

___
*[`src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qSk2Y36rfMpH0DUlV:-O4qmpf31zPE2uTfkBnC:b-k1y5pj) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextBackupTermsCorpus.cs#L17)):*

Good point. I missed that. You can provide the option to pass the settings to the constructor of `ZipParatextTermsParser`. If the settings is null, then you can use the settings parser, similar to what we have in `ParatextProjectTextUpdaterBase`.
</blockquote></details>

OK. Done.
___
*[`src/SIL.Machine/Corpora/ParatextTermsParserBase.cs` line 12 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4q_PUy56VumIi5TrDx:-O4qmjOl275uYv48LrQA:b-6boojw) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextTermsParserBase.cs#L12)):*

You should rename this to `ParatextProjectTermsParserBase` to match the other PT project classes.
</blockquote></details>

Done. + child classes
___
*[`src/SIL.Machine/Corpora/ParatextTermsParserBase.cs` line 37 at r6](https://reviewable.io/reviews/sillsdev/machine/233#-O4qRA0X1NCKtGnMJIJ4:-O4qmilH7qkIW438wAfo:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/31b402ab74f209351d9bee1f0ea471a3c3b6d543/src/SIL.Machine/Corpora/ParatextTermsParserBase.cs#L37)):*

I would make the return type `IEnumerable<(string TermId, IReadOnlyList<string> Glosses)>`.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> Reviewed 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/233#-:-O4r0FOGAuHHpaSICIbJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #235, opened 2024-08-23
Ignore freestanding ellipses

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/458.

I think this is the simplest solution. Let me know what you think.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/235)

### DIFF
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -171,7 +171,7 @@ namespace SIL.Machine.Corpora
         public override void Text(UsfmParserState state, string text)
         {
             // if we hit text in a verse paragraph and we aren't in a verse, then start a non-verse segment
-            if (text.Trim().Length > 0)
+            if (text.Trim().Length > 0 && text.Trim() != "...")
                 CheckConvertVerseParaToNonVerse(state);
         }
 
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -171,7 +171,7 @@ namespace SIL.Machine.Corpora
         public override void Text(UsfmParserState state, string text)
         {
             // if we hit text in a verse paragraph and we aren't in a verse, then start a non-verse segment
-            if (text.Trim().Length > 0 && text.Trim() != "...")
+            if (text.Trim().Length > 0)
                 CheckConvertVerseParaToNonVerse(state);
         }
 
--- a/src/SIL.Machine/Corpora/UsfmTextBase.cs
+++ b/src/SIL.Machine/Corpora/UsfmTextBase.cs
@@ -235,11 +235,6 @@ namespace SIL.Machine.Corpora
                 if (_rowTexts.Count == 0)
                     return;
 
-                if (text.Trim() == "...")
-                {
-                    text = "";
-                }
-
                 StringBuilder rowText = _rowTexts.Peek();
                 if (_text._includeMarkers)
                 {
@@ -279,6 +274,10 @@ namespace SIL.Machine.Corpora
             protected override void EndVerseText(UsfmParserState state, IReadOnlyList<ScriptureRef> scriptureRefs)
             {
                 string text = _rowTexts.Pop().ToString();
+                if (text.Trim() == "...")
+                {
+                    text = "";
+                }
                 _rows.AddRange(_text.CreateRows(scriptureRefs, text, _sentenceStart));
                 _sentenceStart = state.Token.Marker == "c" || text.HasSentenceEnding();
             }
--- a/src/SIL.Machine/Corpora/UsfmTextBase.cs
+++ b/src/SIL.Machine/Corpora/UsfmTextBase.cs
@@ -235,6 +235,11 @@ namespace SIL.Machine.Corpora
                 if (_rowTexts.Count == 0)
                     return;
 
+                if (text.Trim() == "...")
+                {
+                    text = "";
+                }
+
                 StringBuilder rowText = _rowTexts.Peek();
                 if (_text._includeMarkers)
                 {
--- a/tests/SIL.Machine.Tests/Corpora/UsfmMemoryTextTests.cs
+++ b/tests/SIL.Machine.Tests/Corpora/UsfmMemoryTextTests.cs
@@ -184,6 +184,47 @@ description
         });
     }
 
+    [Test]
+    public void GetRows_IgnoreLoneEllipsis()
+    {
+        TextRow[] rows = GetRows(
+            @"\id MAT - Test
+\c 1
+\q1
+\f \fr 119 \ft World \f*
+\v 1 ...
+\v 2 Text
+\c 2
+\d
+description
+\b
+",
+            includeAllText: true
+        );
+        Assert.Multiple(() =>
+        {
+            Assert.That(rows, Has.Length.EqualTo(5), string.Join(",", rows.Select(tr => tr.Text)));
+            Assert.That(rows[1].Text, Is.EqualTo(""));
+        });
+    }
+
+    [Test]
+    public void GetRows_Ellipsis()
+    {
+        TextRow[] rows = GetRows(
+            @"\id MAT - Test
+\c 1
+\v 1 Verse text ... More text
+",
+            includeAllText: true
+        );
+        Assert.Multiple(() =>
+        {
+            Assert.That(rows, Has.Length.EqualTo(1), string.Join(",", rows.Select(tr => tr.Text)));
+            Assert.That(rows[0].Text, Is.EqualTo("Verse text ... More text"));
+        });
+    }
+
     private static TextRow[] GetRows(string usfm, bool includeMarkers = false, bool includeAllText = false)
     {
         UsfmMemoryText text =


johnml1135:
*[`tests/SIL.Machine.Tests/Corpora/UsfmMemoryTextTests.cs` line 217 at r1](https://reviewable.io/reviews/sillsdev/machine/235#-O5-QiKl9C63ZbWusO1G:-O5-QiKl9C63ZbWusO1H:bbn0dvy) ([raw file](https://github.com/sillsdev/machine/blob/7f026091c19339559fb58517218ed649b620e4b7/tests/SIL.Machine.Tests/Corpora/UsfmMemoryTextTests.cs#L217)):*
> ```Smalltalk
>             @"\id MAT - Test
> \c 1
> \v 1 Verse text ... More text
> ```

What about things like this:
```
\v 1 ...
\q1 ...
\q2 ...
\v2 ... \f footnote \f*
```

I'm probably doing it wrong, but what is the defined behavior if the ellipses is between other USFM markers?

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 241 at r1](https://reviewable.io/reviews/sillsdev/machine/235#-O5-RU8v7s7D52IVbOyn:-O5-RU8wB0ql5yBzZQ0b:b-ix22hj) ([raw file](https://github.com/sillsdev/machine/blob/7f026091c19339559fb58517218ed649b620e4b7/src/SIL.Machine/Corpora/UsfmTextBase.cs#L241)):*
> ```Smalltalk
>                 {
>                     text = "";
>                 }
> ```

Should this happen before tokenization?  I'm not quite understanding.

johnml1135:
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 174 at r1](https://reviewable.io/reviews/sillsdev/machine/235#-O5-RxeWDxYSUAWGOkxY:-O5-RxeWDxYSUAWGOkxZ:b-kejet9) ([raw file](https://github.com/sillsdev/machine/blob/7f026091c19339559fb58517218ed649b620e4b7/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L174)):*
> ```Smalltalk
>         {
>             // if we hit text in a verse paragraph and we aren't in a verse, then start a non-verse segment
>             if (text.Trim().Length > 0 && text.Trim() != "...")
> ```

This seems like the wrong place to change it - this was added to solve a different issue.  Does this change cause the tests to fail?  It likely should...

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 241 at r1](https://reviewable.io/reviews/sillsdev/machine/235#-O5-RU8v7s7D52IVbOyn:-O5-WdFX7AZN3x5dA2QJ:bpfhxkm) ([raw file](https://github.com/sillsdev/machine/blob/7f026091c19339559fb58517218ed649b620e4b7/src/SIL.Machine/Corpora/UsfmTextBase.cs#L241)):*

Should this happen before tokenization?  I'm not quite understanding.
</blockquote></details>

This may work - but I don't understand this part of the parse to know where all the places that it will work.  I also don't know how paratext populates \q1, and other poetry markers - nor to I know how this may handle populating section headers etc. with ... if that is an issue.

johnml1135 (commented):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Enkidu93)

Enkidu93 (commented):
To respond to all your comments in one place, @johnml1135, my goal was to treat verse's whose only text is '...' the same as white-space-only verses. This might be a hacky solution, but I'm not sure what the less hacky solution would be. I could certainly define a regex or something if we're interested in potentially treating other text as white space (e.g., only punctuation or something). Or maybe there's a way to think of this as a special kind of usfm token, but I also don't want to go overboard just for this small issue.

Judging from the project I was working with in which I originally saw this issue, it appears that '...' only appears as filler for an entire verse text - i.e., I didn't see examples of multiple poetic lines (or something similar) each with '...', but I'm not sure if those might exist as well since, like I said, I'm just basing this off one project.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5Dv2tx8p3MGxazqZeD:b-wiy23z)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

ddaspit (changes requested):
We are trying to handle the specific case where Paratext inserts ellipses to fill an empty verse. I think you should narrow the check even further and only apply it when you are creating text rows for a verse, i.e. `UsfmTextBase.EndVerseText`. I want to avoid incorrectly interpreting the intentional use of ellipses as empty text. 

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5EEbWE2h871_z1_fk5:bcchtqr)* status: all files reviewed, 3 unresolved discussions (waiting on @Enkidu93)

johnml1135:
I agree that changing the code there would prevent partial matches.  @ddaspit - do you know all the instances in which the ellipsis might be added in Paratext?

Enkidu93 (commented):
Like this, @ddaspit ? 

Yeah, it would be nice to know what Paratext's behavior is such that we can make appropriately narrow and also so that we can have reasonable tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5EZTyi0FCXAu6KOt1b:b-jr0hi9)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

johnml1135:
*[`src/SIL.Machine/Corpora/UsfmTextBase.cs` line 276 at r2](https://reviewable.io/reviews/sillsdev/machine/235#-O5EbeSP2Dw1XNTHUHjU:-O5EbeSP2Dw1XNTHUHjV:b-uqhsxh) ([raw file](https://github.com/sillsdev/machine/blob/0d743a7c13a9c17bb99bd1626d5c434a70158b57/src/SIL.Machine/Corpora/UsfmTextBase.cs#L276)):*
> ```Smalltalk
>             protected override void EndVerseText(UsfmParserState state, IReadOnlyList<ScriptureRef> scriptureRefs)
>             {
>                 string text = _rowTexts.Pop().ToString();
> ```

I think a simple comment would be appropriate here - what are we doing and why?  "Paratext sometimes auto-adds an ellipses for newly instantiated verses.  This should be ignored for NLP tasks." or similar.

johnml1135 (commented):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

ddaspit (commented):
Let me check on the Paratext behavior and get back to you.

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5FB8VwDafndW-5ksyW:bo1amjn)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

ddaspit (commented):
I can't figure out where the ellipses are coming from. I thought they were something that Paratext would insert in some cases, but I can't find any evidence of that. They might be coming from [Adapt It](https://adapt-it.org/Adapt_It_Quick_Start.htm#Inserting_null_source_phrase), but I'm not sure how the ellipses would get back into the source USFM from Adapt It. If this isn't standard Paratext behavior, we shouldn't handle this in Machine. Instead, we should handle this during preprocessing in Serval.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5FfJ1T4qQjypCmt9Lc:bdnzvcg)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93 (commented):
Maybe we can ask Michael about it tomorrow (?), @ddaspit, before committing to that switch. He seemed fairly sure that Paratext created the ellipses when the team made an empty book and my impression was that this was typical. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5Fv2Os0W9BJzube5q8:b5diz2a)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

ddaspit (commented):
I spoke to Michael about it. He doesn't know where they come from. I actually looked at the Paratext repo and found no trace of any code that adds ellipses. I looked at a couple of projects that contain them and can't find a consistent pattern.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5G5-2W8mis4sMxbWuT:bwvbglc)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93 (commented):
OK, so what's the plan here. Do you want to relocate this logic? What do you mean by preprocessing in Serval? There isn't really any file preprocessing is there, right?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5Jziys0hLDkxe2BfLH:bz4w1l6)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

ddaspit (commented):
We want to filter out any segments that are `...` when generating the training corpus and pretranslation JSON file in  `PreprocessBuildJob.WriteDataFilesAsync`. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5K50atAaVCT1j-OZWN:b-192p7b)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93 (commented):
OK, just making sure I wasn't missing something :laughing: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/235#-:-O5KOUoPCjDlGKUtcFVl:b-as2oyp)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

PR #259, opened 2024-10-17
LT-21911: Fix problem that suppressed homographs

jtmaxwell3:
Beth reported that the novel root guesser couldn't handle homographs (e.g. two lexical entries that looked like [Seg]*).  This was because there was code to suppress duplicate paths through a pattern, but the scope was too big.  Narrowing the scope fixed the problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/259)


ddaspit (commented):
Does this need a unit test?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/259#-:-O9Q99BH-9GIZ2MnO9qJ:bs29kai)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

jtmaxwell3:
I don't think that it needs a unit test.  It was just a bug in how I wrote the code, and I don't think that anybody else would make the same mistake.  But I can add a unit test if you want.

PR #282, opened 2025-01-21
Merge duplicate analyses

jtmaxwell3:
This merges equivalent analyses in AnalysisStratumRule.Apply and then expands alternatives after they have been found in the lexicon in Morpher.  This can speed things up by 3-4x if there are multiple strata since fewer analyses get passed from one stratum to the next.  It required adding a Source field and an Alternatives field to Word so that the alternatives can be reconstructed by Word.ExpandAlternatives.  The Source field gets filled in with the original word when a Word is cloned so that past alternatives can be accessed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/282)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs
@@ -67,6 +67,7 @@ namespace SIL.Machine.Morphology.HermitCrab
             if (_morpher.TraceManager.IsTracing)
                 _morpher.TraceManager.BeginUnapplyStratum(_stratum, input);
 
+            Word origInput = input;
             input = input.Clone();
             input.Stratum = _stratum;
 
@@ -89,11 +90,13 @@ namespace SIL.Machine.Morphology.HermitCrab
             {
                 if (mergeEquivalentAnalyses)
                 {
+                    // Skip intermediate sources from phonological rules, templates, and morphological rules.
+                    mruleOutWord.Source = origInput;
                     Shape shape = mruleOutWord.Shape;
                     if (shapeWord.ContainsKey(shape))
                     {
                         shapeWord[shape].Alternatives.Add(mruleOutWord);
-                         continue;
+                        continue;
                     }
                     shapeWord[shape] = mruleOutWord;
                 }
--- a/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs
@@ -72,6 +72,11 @@ namespace SIL.Machine.Morphology.HermitCrab
 
             _prulesRule.Apply(input);
             input.Freeze();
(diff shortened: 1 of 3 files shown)


ddaspit (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/282)* status: all files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

___
*[`src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs` line 96 at r2](https://reviewable.io/reviews/sillsdev/machine/282#-OHETBiH6MKNDQVJpR3H:-OHETBiIAF6EaZOYtWqN:b7xicin) ([raw file](https://github.com/sillsdev/machine/blob/1194245454bf566caba4f179785ca84a893f4d9e/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs#L96)):*
> ```Smalltalk
>                 {
>                     Shape shape = mruleOutWord.Shape;
>                     if (shapeWord.ContainsKey(shape))
> ```

If you use `TryGetValue`, it will only query the dictionary once.

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/282)* status: all files reviewed, 2 unresolved discussions (waiting on @jtmaxwell3)

___
*[`src/SIL.Machine.Morphology.HermitCrab/Word.cs` line 77 at r2](https://reviewable.io/reviews/sillsdev/machine/282#-OHEUZIv0sn_wX4eMdVQ:-OHEUZIv0sn_wX4eMdVR:bgweubt) ([raw file](https://github.com/sillsdev/machine/blob/1194245454bf566caba4f179785ca84a893f4d9e/src/SIL.Machine.Morphology.HermitCrab/Word.cs#L77)):*
> ```Smalltalk
>             Stratum = word.Stratum;
>             Source = word;
>             // Don't copy Alternatives.
> ```

I'm not sure if I totally understand the purpose of `Source`, but could you do a shallow copy of the alternatives instead of using the `Source` property?

jtmaxwell3:
You could do a shallow copy of the alternatives, but then you would need to reconstruct the alternatives at the time of the clone (e.g. do what ExpandAlternatives does).  By leaving the alternatives in place and storing Source, we can avoid calling ExpandAlternatives unless the final word is in the lexicon.  This is what improves performance.  The Source property creates a chain of words that represents the history of the word's processing.  The alternatives are stored in the chain in multiple places.  ExpandAlternatives only reconstructs alternatives for some of the words.  I don't think that it should be part of the code for clone words.

ddaspit (approved):
Thanks for the explanation.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/282#-:-OHEbqKV3wp_kj2qtaqb:b-wsvas3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

Issue #294, opened 2025-03-17
`NParallelTextCorpus` does not properly handle successive ranges where the target has empty segments


Enkidu93:
This is the root issue behind an issue @pmachapman reported where the content of multiple ranges were getting concatenated.

PR #295, opened 2025-03-17
Fix faulty edge-case logic in NParallelTextCorpus

Enkidu93:
Fixes https://github.com/sillsdev/machine/issues/294

This was a bug that already existed in the previous `ParallelTextCorpus` implementation. We were relying on segment count as an indicator of a new range starting when we should just have used the `IsRangeStart` property since this isn't reliable when your target is totally empty.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/295)

### DIFF
--- a/src/SIL.Machine/Corpora/NParallelTextCorpus.cs
+++ b/src/SIL.Machine/Corpora/NParallelTextCorpus.cs
@@ -71,9 +71,9 @@ namespace SIL.Machine.Corpora
             }
         }
 
-        private static bool AllInRangeHaveSegments(IList<TextRow> rows)
+        private static bool AllRangesAreNewRanges(IList<TextRow> rows)
         {
-            return rows.All(r => (r.IsInRange && !r.IsEmpty) || (!r.IsInRange));
+            return rows.All(r => r.IsRangeStart || !r.IsInRange);
         }
 
         private IList<int> MinRefIndexes(IList<object> refs)
@@ -211,7 +211,7 @@ namespace SIL.Machine.Corpora
                     {
                         if (
                             rangeInfo.IsInRange
-                            && AllInRangeHaveSegments(currentRows.Where((r, i) => !completed[i]).ToArray())
+                            && AllRangesAreNewRanges(currentRows.Where((r, i) => !completed[i]).ToArray())
                         )
                         {
                             yield return rangeInfo.CreateRow();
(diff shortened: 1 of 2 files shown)


johnml1135:
*[`tests/SIL.Machine.Tests/Corpora/ParallelTextCorpusTests.cs` line 604 at r1](https://reviewable.io/reviews/sillsdev/machine/295#-OLZyQgF7pmkhceP_SlI:-OLZyQgF7pmkhceP_SlJ:b-djqtdd) ([raw file](https://github.com/sillsdev/machine/blob/d852fbd947b283929d57b9ca5d43a10dfa9e289e/tests/SIL.Machine.Tests/Corpora/ParallelTextCorpusTests.cs#L604)):*
> ```Smalltalk
>     [Test]
>     public void GetRows_AdjacentRangesBothTexts_EmptyTargetSegments()
>     {
> ```

I'm going to have to assume that this test catches the condition and would fail without the code changes. Is that correct and has that been verified?

johnml1135 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/295)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @Enkidu93)

Enkidu93:
> _[`tests/SIL.Machine.Tests/Corpora/ParallelTextCorpusTests.cs` line 604 at r1](https://reviewable.io/reviews/sillsdev/machine/295#-OLZyQgF7pmkhceP_SlI:-OLZyQgF7pmkhceP_SlJ:b-djqtdd) ([raw file](https://github.com/sillsdev/machine/blob/d852fbd947b283929d57b9ca5d43a10dfa9e289e/tests/SIL.Machine.Tests/Corpora/ParallelTextCorpusTests.cs#L604)):_
> 
> > ```smalltalk
> >     [Test]
> >     public void GetRows_AdjacentRangesBothTexts_EmptyTargetSegments()
> >     {
> > ```
> 
> I'm going to have to assume that this test catches the condition and would fail without the code changes. Is that correct and has that been verified?

Yes. I just re-verified. It exhibits the behavior Peter described when the code is reverted causing the test to fail.

johnml1135 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/295)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #296, opened 2025-03-17
Fix LT-21867: Optional contexts don't work properly in Hermit Crab

jtmaxwell3:
https://jira.sil.org/browse/LT-21867 reports that optional contexts don't work correctly.  If a context is optional, then it should accept anything.  But the code to traverse fsts doesn't accept the empty string.  I wasn't able to fix the code to traverse fsts, so I changed the context code to check whether the context accepts the empty string.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/296)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/RewriteRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/RewriteRuleSpec.cs
@@ -99,7 +99,7 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
                         rightEnvMatch == null
                         || rightEnvMatch.Success
                         || subruleSpec.RightEnvironmentMatcher.AcceptsEmpty()
-                       )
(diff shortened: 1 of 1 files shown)


ddaspit (commented):
This is definitely a bug in the `Fst` class. The class should properly handle the case where the start state is accepting. I would prefer that the `Fst` class is fixed. I took a stab at getting it to work in [this PR](https://github.com/sillsdev/machine/pull/298). Let me know if it fixes the issue.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/296#-:-OLpaXj4DQxggFzZb8Qh:b-t54cxs)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

jtmaxwell3:
Fixing the Fst class is a better solution, and it seems to work.  Thanks!  You can delete this pull request.

PR #306, opened 2025-05-30
Port block handler

Enkidu93:
@ddaspit, when porting this, is speed a concern? I've tried to follow machine.py as closely as possible with few exceptions, but there's definitely room for optimization. Is this something we should think about now or just something we should revisit later on?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/306)

### DIFF
--- a/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs
+++ b/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs
@@ -1,0 +1,360 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using SIL.Extensions;
+using SIL.Machine.Translation;
+
+namespace SIL.Machine.Corpora
+{
+    public class PlaceMarkersAlignmentInfo
+    {
+        public IReadOnlyList<string> Refs { get; }
+        public IReadOnlyList<string> SourceTokens { get; }
+        public IReadOnlyList<string> TargetTokens { get; }
+        public WordAlignmentMatrix Alignment { get; }
+
+        public PlaceMarkersAlignmentInfo(
+            IReadOnlyList<string> refs,
+            IReadOnlyList<string> sourceTokens,
+            IReadOnlyList<string> targetTokens,
+            WordAlignmentMatrix alignment
+        )
+        {
+            Refs = refs;
+            SourceTokens = sourceTokens;
+            TargetTokens = targetTokens;
+            Alignment = alignment;
+        }
+    }
+
+    public class PlaceMarkersUsfmUpdateBlockHandler : IUsfmUpdateBlockHandler
+    {
+        private readonly IDictionary<string, PlaceMarkersAlignmentInfo> _alignmentInfo;
+
+        public PlaceMarkersUsfmUpdateBlockHandler(IEnumerable<PlaceMarkersAlignmentInfo> alignmentInfo)
+        {
+            _alignmentInfo = alignmentInfo.ToDictionary(info => info.Refs.First(), info => info);
+        }
+
+        public UsfmUpdateBlock ProcessBlock(UsfmUpdateBlock block)
+        {
+            string reference = block.Refs.FirstOrDefault().ToString();
+            List<UsfmUpdateBlockElement> elements = block.Elements.ToList();
+
+            // Nothing to do if there are no markers to place or no alignment to use
+            if (
+                elements.Count == 0
+                || !_alignmentInfo.TryGetValue(reference, out PlaceMarkersAlignmentInfo alignmentInfo)
+                || alignmentInfo.Alignment.RowCount == 0
+                || alignmentInfo.Alignment.ColumnCount == 0
+                || !elements.Any(e =>
+                    e.Type.IsOneOf(UsfmUpdateBlockElementType.Paragraph, UsfmUpdateBlockElementType.Style)
+                    && !e.MarkedForRemoval
+                )
+            )
+            {
+                return block;
+            }
+
+            // Paragraph markers at the end of the block should stay there
+            // Section headers should be ignored but re-inserted in the same position relative to other paragraph markers
+            List<UsfmUpdateBlockElement> endElements = new List<UsfmUpdateBlockElement>();
+            bool eobEmptyParas = true;
+            List<(int ParaMarkersLeft, UsfmUpdateBlockElement Element)> headerElements =
+                new List<(int paraMarkersLeft, UsfmUpdateBlockElement element)>();
+            int paraMarkersLeft = 0;
+            foreach ((int i, UsfmUpdateBlockElement element) in elements.Select((e, i) => (i, e)).Reverse())
+            {
+                if (element.Type == UsfmUpdateBlockElementType.Paragraph && !element.MarkedForRemoval)
+                {
+                    if (element.Tokens.Count > 1)
+                    {
+                        headerElements.Insert(0, (paraMarkersLeft, element));
+                        elements.RemoveAt(i);
+                    }
+                    else
+                    {
+                        paraMarkersLeft++;
+
+                        if (eobEmptyParas)
+                        {
+                            endElements.Insert(0, element);
+                        }
+                    }
+                }
+                else if (
+                    !(
+                        element.Type == UsfmUpdateBlockElementType.Embed
+                        || (
+                            element.Type == UsfmUpdateBlockElementType.Text
+                            && element.Tokens[0].ToUsfm().Trim().Count() == 0
+                        )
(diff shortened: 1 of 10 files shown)


ddaspit (changes requested):
> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/306)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 61 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4xDRyDzeCGLKQMEKh:-OS4xDRyDzeCGLKQMEKi:b-7ml68m) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L61)):*
> ```Smalltalk
>             // Paragraph markers at the end of the block should stay there
>             // Section headers should be ignored but re-inserted in the same position relative to other paragraph markers
>             List<UsfmUpdateBlockElement> endElements = new List<UsfmUpdateBlockElement>();
> ```

You can use `var` for all of these variable declarations, since the type is explicit on the right hand side.
___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 91 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4wucmA-3VeCl-Jk8v:-OS4wucmA-3VeCl-Jk8w:bks93cr) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L91)):*
> ```Smalltalk
>                         || (
>                             element.Type == UsfmUpdateBlockElementType.Text
>                             && element.Tokens[0].ToUsfm().Trim().Count() == 0
> ```

You should use `Length` instead of the LINQ `Count` function.
___
*[`tests/SIL.Machine.Tests/Corpora/PlaceMarkersUsfmUpdateBlockHandlerTests.cs` line 10 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4xio2C_WZVlm1wIin:-OS4xio39XNSSyLWSIRK:bi1htpn) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/tests/SIL.Machine.Tests/Corpora/PlaceMarkersUsfmUpdateBlockHandlerTests.cs#L10)):*
> ```Smalltalk
> public class PlaceMarkersUsfmUpdateBlockHandlerTests
> {
>     private LatinWordTokenizer Tokenizer { get; set; }
> ```

This can probably just be a static read-only field.
___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 535 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4vvdR7R_Hl02jgDk_:-OS4vvdR7R_Hl02jgDka:b-5523iv) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L535)):*
> ```Smalltalk
>             }
>             List<UsfmToken> tokens = updateBlock.GetTokens();
>             paraElems.Reverse();
> ```

It is a small thing, but I would just use the LINQ `Reverse` function in the `foreach` loop instead of the in-place `Reverse` method on `List`. It avoids an extra iteration of the list. You can call the LINQ reverse like this `Enumerable.Reverse(paraElems)`.

ddaspit (commented):
@Enkidu93 What kind of optimizations are you talking about?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/306#-:-OS5PueV5H0Rs7Il2BXl:b-59qoc7)* status: all files reviewed, 4 unresolved discussions (waiting on @Enkidu93)

Enkidu93 (commented):
I'm seeing lots of `Insert(index)` and `RemoveAt` and `IndexOf`+`Substring`'s as well as what looks like iterating through collections more than once. These make suspicious that we could improve the performance. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/306#-:-OS5aU9dD93x4lIOiGf1:b2zsga0)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 61 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4xDRyDzeCGLKQMEKh:-OS5XkPt3ChMUGhJg3GV:blryxv5) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L61)):*

You can use `var` for all of these variable declarations, since the type is explicit on the right hand side.
</blockquote></details>

Done. I thought we were using the `Type name = new()` syntax and avoiding `var`?
___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 91 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4wucmA-3VeCl-Jk8v:-OS5Xp8k3QHw5nsAOJiJ:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L91)):*

You should use `Length` instead of the LINQ `Count` function.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 535 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4vvdR7R_Hl02jgDk_:-OS5ZV8_9C1PqQIz1FfB:bjrsjfc) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L535)):*

It is a small thing, but I would just use the LINQ `Reverse` function in the `foreach` loop instead of the in-place `Reverse` method on `List`. It avoids an extra iteration of the list. You can call the LINQ reverse like this `Enumerable.Reverse(paraElems)`.
</blockquote></details>

Oooo, thank you! I only didn't do this because I didn't know it existed and was being stupid :laughing:. That makes sense since you can call `Select().Reverse()`. Done. 
___
*[`tests/SIL.Machine.Tests/Corpora/PlaceMarkersUsfmUpdateBlockHandlerTests.cs` line 10 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4xio2C_WZVlm1wIin:-OS5aSteE6mvNOCp3rnG:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/tests/SIL.Machine.Tests/Corpora/PlaceMarkersUsfmUpdateBlockHandlerTests.cs#L10)):*

This can probably just be a static read-only field.
</blockquote></details>

Done.

Enkidu93 (commented):
After looking through it again, I can definitely see room for optimizations but it would require enough reworking (which should then probably be back-ported to machine.py) that I've just spun off an issue for now in the interest of meeting the quarter deadline. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/306#-:-OSK4ciODnY44BK_8g08:btf8kll)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 61 at r4](https://reviewable.io/reviews/sillsdev/machine/306#-OS4xDRyDzeCGLKQMEKh:-OSK3waA5Tei4u15IhEg:b9degaa) ([raw file](https://github.com/sillsdev/machine/blob/5495c05e2afc3d155e6fd14ff8812603f2d5eaa9/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L61)):*

I wasn't sure if we could use that syntax in this assembly, since it targets .NET Standard 2.0, which is locked to an older version of C#.
</blockquote></details>

Yeah :thumbsup:. I wasn't sure if when we couldn't use the `new()` syntax, if you preferred we avoid  `var`.

PR #307, opened 2025-06-02
Improve error handling for Hermit Crab users

jtmaxwell3:
I wrapped exceptions in the Hermit Crab compiler so that the user would know which rule couldn't be compiled.  This may allow the user to fix the rule causing the problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/307)


jtmaxwell3:
@ddaspit I added a CompileException class, as you requested.

PR #313, opened 2025-06-23
Fix for LT-18767 Yellow crash foe XAmple-style redup pattern

AndyBlack:
When using an XAmple-style reduplication pattern, if there was an indexed natural class in the pattern that was not also in the environment, then Hermit Crab threw a key not found exception.  This by-passes that.

Note: I wanted to create a unit test case for this but did not know how to create the reduplication information in an Allomorph.  So there is no unit test case for this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/313)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/CopyFromInput.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/CopyFromInput.cs
@@ -18,6 +18,14 @@ namespace SIL.Machine.Morphology.HermitCrab.MorphologicalRules
             IDictionary<string, int> capturedParts
         )
         {
+            if (!partLookup.ContainsKey(PartName))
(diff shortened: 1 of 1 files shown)


ddaspit (changes requested):
It would be good if we had a unit test that covers this case.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/313#-:-OTT4uHm25cWk54OBgiU:b-mwgqwz)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AndyBlack)

ddaspit (commented):
Oops, I missed that. I will see what I can do.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/313#-:-OTThjIcC8IWZRJhSl5N:b70dae1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AndyBlack)

PR #322, opened 2025-08-18
LT-22190 Use BitArray for large symbolic feature value sets

AndyBlack:
This is a fix related to "L[T-22190 Hermit Crab adds wrong POSes to compound rules when there are more than 64 POSes](url)"

The problem is that Machine uses a ulong to represent the possible values of a feature.  This solution uses a BitArray instead of a ulong whenever there are more than sizeof(ulong) * 8 (=64) values.

The solution has a subclass of SymbolicFeatureValue called SymbolicFeatureValueBA which is used whenever there are more than 64 possible symbol values.  It adds a SymbolicFeatureValueFactory singleton class to create symbolic feature value objects.

The SymbolicFeatureValue class also now has a static method called GetFeatureSymbolFromFeatureStruct().  For reasons not clear to me, the static implicit operator SymbolicFeatureValue(FeatureSymbol symbol) method returns a null when invoked on a SymbolicFeatureValueBA oject.  This GetFeatureSymbolFromFeatureStruct() method is a way I found to make things work.  If there is a better way to do it, please let me know.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/322)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/CharacterDefinition.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/CharacterDefinition.cs
@@ -19,7 +19,11 @@ namespace SIL.Machine.Morphology.HermitCrab
 
         public FeatureSymbol Type
         {
-            get { return (FeatureSymbol)_fs.GetValue(HCFeatureSystem.Type); }
+            get
+            {
+                FeatureSymbol fsym = SymbolicFeatureValue.GetFeatureSymbolFromFeatureStruct(_fs, HCFeatureSystem.Type);
+                return fsym;
+            }
         }
 
         public ReadOnlyCollection<string> Representations
--- a/src/SIL.Machine.Morphology.HermitCrab/HCFeatureSystem.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/HCFeatureSystem.cs
@@ -51,7 +51,7 @@ namespace SIL.Machine.Morphology.HermitCrab
             Modified = new SymbolicFeature(Guid.NewGuid().ToString(), Dirty, Clean)
             {
                 Description = "Modified",
-                DefaultValue = new SymbolicFeatureValue(Clean)
+                DefaultValue = SymbolicFeatureValueFactory.Instance.Create(Clean)
             };
 
             Deleted = new FeatureSymbol(Guid.NewGuid().ToString()) { Description = "Deleted" };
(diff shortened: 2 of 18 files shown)


ddaspit (changes requested):
> @ddaspit reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/322)* status: 0 of 18 files reviewed, 2 unresolved discussions

___
*[`src/SIL.Machine/FeatureModel/SymbolicFeatureValueFactory.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/machine/322#-OY34zsM48JrSIHgLMLZ:-OY34zsM48JrSIHgLML_:bog98y0) ([raw file](https://github.com/sillsdev/machine/blob/e9efd2bd8ac27b25a00f4b514a5ca65522cb9cd9/src/SIL.Machine/FeatureModel/SymbolicFeatureValueFactory.cs#L7)):*
> ```Smalltalk
> namespace SIL.Machine.FeatureModel
> {
>     public class SymbolicFeatureValueFactory
> ```

Would it be possible to entirely encapsulate the different implementations of symbolic feature value within the `SymbolicFeatureValue` class instead of this new factory? You could use composition rather than inheritance to support the implementations. You could define an interface/ABC that abstracts the two implementations. The constructor for `SymbolicFeatureValue` would create the correct instance and delegate calls to the instance where necessary. This design should greatly reduce the number of changes that are required to the rest of parser.
___
*[`src/SIL.Machine/FeatureModel/SymbolicFeatureValueBA.cs` line 221 at r1](https://reviewable.io/reviews/sillsdev/machine/322#-OY378Fe2mgQcqCnwi2O:-OY378Fe2mgQcqCnwi2P:b-eh4cjh) ([raw file](https://github.com/sillsdev/machine/blob/e9efd2bd8ac27b25a00f4b514a5ca65522cb9cd9/src/SIL.Machine/FeatureModel/SymbolicFeatureValueBA.cs#L221)):*
> ```Smalltalk
>             {
>                 BitArray notOtherFlags = ((BitArray)otherFlags.Clone()).Not();
>                 BitArray notOtherFlagsAndFeatureMask = new BitArray(notOtherFlags.And(Feature.MaskBA));
> ```

It feels like the number of BitArray instances could be reduced. Could we reuse some of the `BitArray` instances?

AndyBlack:
On the BitArray instances: Unlike with ulong, applying any operation to a BitArray changes its value.  I ran into failed tests because of one being negated, say.  Using Clone() or creating a new instance got around this problem.

AndyBlack:
This is replaced by https://github.com/sillsdev/machine/pull/330

Issue #338, opened 2025-09-22
Inconsistent comparison of verses results in missing verses


Enkidu93:
Pretranslation rows are sorted by `ScriptureRef.CompareTo(compareSegments: true)` but `UpdateUsfmParserHandler` uses `ScriptureRef.CompareTo(compareSegments: false)` which can result in missing verses.

For example,  for row refs  `["ISA 9:1a", "ISA 9:1a/1:s1", "ISA 9:1b", "ISA 9:2"]`, updating the USFM for `"ISA"` results in verse `"9:1b"` missing.

Enkidu93:
If we use the strategy suggested [here](https://github.com/sillsdev/machine.py/issues/232), we may not need to solve this by implementing a `ScriptureRef` comparer that ignores segments.

PR #341, opened 2025-09-24
Get update rows by reference

Enkidu93:
Fixes https://github.com/sillsdev/machine.py/issues/232
Also fixes https://github.com/sillsdev/machine/issues/339 and fixes https://github.com/sillsdev/machine/issues/340

These will all need to be ported.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/341)

### DIFF
--- a/src/SIL.Machine/Corpora/ParatextProjectTextUpdaterBase.cs
+++ b/src/SIL.Machine/Corpora/ParatextProjectTextUpdaterBase.cs
@@ -29,7 +29,8 @@ namespace SIL.Machine.Corpora
             UpdateUsfmMarkerBehavior styleBehavior = UpdateUsfmMarkerBehavior.Strip,
             IEnumerable<string> preserveParagraphStyles = null,
             IEnumerable<IUsfmUpdateBlockHandler> updateBlockHandlers = null,
-            IEnumerable<string> remarks = null
+            IEnumerable<string> remarks = null,
+            bool compareSegments = false
         )
         {
             string fileName = _settings.GetBookFileName(bookId);
@@ -51,7 +52,8 @@ namespace SIL.Machine.Corpora
                 styleBehavior,
                 preserveParagraphStyles,
                 updateBlockHandlers,
-                remarks
+                remarks,
+                compareSegments
             );
             try
             {
--- a/src/SIL.Machine/Corpora/ParatextProjectTextUpdaterBase.cs
+++ b/src/SIL.Machine/Corpora/ParatextProjectTextUpdaterBase.cs
@@ -30,6 +30,7 @@ namespace SIL.Machine.Corpora
             IEnumerable<string> preserveParagraphStyles = null,
             IEnumerable<IUsfmUpdateBlockHandler> updateBlockHandlers = null,
             IEnumerable<string> remarks = null,
+            Func<UsfmUpdateBlockHandlerException, bool> errorHandler = null,
             bool compareSegments = false
         )
         {
@@ -53,6 +54,7 @@ namespace SIL.Machine.Corpora
                 preserveParagraphStyles,
                 updateBlockHandlers,
                 remarks,
+                errorHandler,
                 compareSegments
             );
             try
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -130,38 +130,4 @@ namespace SIL.Machine.Corpora
             return sb.ToString();
         }
     }
-
-    public class ScriptureRefComparer : IComparer<ScriptureRef>, IEqualityComparer<ScriptureRef>
-    {
-        public static ScriptureRefComparer Default { get; } = new ScriptureRefComparer(compareSegments: true);
-        public static ScriptureRefComparer IgnoreSegments { get; } = new ScriptureRefComparer(compareSegments: false);
-        private readonly bool _compareSegments;
-
-        public ScriptureRefComparer(bool compareSegments = true)
-        {
-            _compareSegments = compareSegments;
-        }
-
-        public int Compare(ScriptureRef x, ScriptureRef y)
-        {
-            return x.CompareTo(y, _compareSegments);
-        }
-
-        public bool Equals(ScriptureRef x, ScriptureRef y)
-        {
-            return x.CompareTo(y, _compareSegments) == 0;
-        }
-
-        public int GetHashCode(ScriptureRef obj)
-        {
-            int hashCode = 23;
-            hashCode =
-                hashCode * 31
-                + (_compareSegments ? obj.VerseRef.BBBCCCVVVS.GetHashCode() : obj.VerseRef.BBBCCCVVV.GetHashCode());
-            hashCode = hashCode * 31 + obj.Versification.GetHashCode();
-            // Using ToRelaxed is necessary to maintain equality across relaxed refs, Equals properly handles relaxed ref comparison
-            hashCode = hashCode * 31 + obj.ToRelaxed().Path.GetSequenceHashCode();
-            return hashCode;
-        }
-    }
 }
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -130,4 +130,38 @@ namespace SIL.Machine.Corpora
             return sb.ToString();
         }
     }
+
+    public class ScriptureRefComparer : IComparer<ScriptureRef>, IEqualityComparer<ScriptureRef>
+    {
+        public static ScriptureRefComparer Default { get; } = new ScriptureRefComparer(compareSegments: true);
+        public static ScriptureRefComparer IgnoreSegments { get; } = new ScriptureRefComparer(compareSegments: false);
+        private readonly bool _compareSegments;
+
+        public ScriptureRefComparer(bool compareSegments = true)
+        {
+            _compareSegments = compareSegments;
+        }
+
+        public int Compare(ScriptureRef x, ScriptureRef y)
+        {
+            return x.CompareTo(y, _compareSegments);
+        }
+
+        public bool Equals(ScriptureRef x, ScriptureRef y)
+        {
+            return x.CompareTo(y, _compareSegments) == 0;
+        }
+
+        public int GetHashCode(ScriptureRef obj)
+        {
+            int hashCode = 23;
+            hashCode =
+                hashCode * 31
+                + (_compareSegments ? obj.VerseRef.BBBCCCVVVS.GetHashCode() : obj.VerseRef.BBBCCCVVV.GetHashCode());
+            hashCode = hashCode * 31 + obj.Versification.GetHashCode();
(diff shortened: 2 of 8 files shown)


Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 76 at r2](https://reviewable.io/reviews/sillsdev/machine/341#-O_wzG7O0etEqPRoabAS:-O_wzG7O0etEqPRoabAT:b-m0eqzh) ([raw file](https://github.com/sillsdev/machine/blob/96253f47f152ba689d12cf1d31f9c84fafb79986/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L76)):*
> ```Smalltalk
>         {
>             // We need two maps so that update rows can be specified per segment
>             // but be handled correctly whether or not the USFM has segments for that verse
> ```

As an alternative to two maps, we could use the ignore segments map and then search through the list to see if a segment matches if the update row ref has a segment. That would probably be better, but I wasn't sure. 
___
*[`tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs` line 520 at r2](https://reviewable.io/reviews/sillsdev/machine/341#-O_wzZ3z2C-WUkI_SdVU:-O_wzZ4-18SLVHnWN7S1:b-ysj3sc) ([raw file](https://github.com/sillsdev/machine/blob/96253f47f152ba689d12cf1d31f9c84fafb79986/tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs#L520)):*
> ```Smalltalk
>             )
>         );
>         // Updating using relaxed refs will not be perfect, but it's the best we can do while allowing for out of order rows
> ```

Are you ok with this, @ddaspit? I don't see a way to avoid this if the rows have relaxed refs.  
___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 134 at r2](https://reviewable.io/reviews/sillsdev/machine/341#-O_wzjm48YtdL8GJO3ml:-O_wzjm52sA1UpAw4wlf:b-7s8kmc) ([raw file](https://github.com/sillsdev/machine/blob/96253f47f152ba689d12cf1d31f9c84fafb79986/src/SIL.Machine/Corpora/ScriptureRef.cs#L134)):*
> ```Smalltalk
>     }
> 
>     public class ScriptureRefComparer : IComparer<ScriptureRef>, IEqualityComparer<ScriptureRef>
> ```

I implemented `IComparer` as well partially just for completeness. There's no way in calling code to compare scripture refs ignoring segments using a comparer (e.g., if you want to use `Order()`) - this was the root isssue of Serval/Machine not handling certain refs correctly. It's superseded by the row map solution, but I still thought it'd be helpful to expose. If you'd rather I remove it, I can.

Enkidu93 (commented):
The changes look great, @ddaspit. Thank you! I think [this issue](https://github.com/sillsdev/machine/issues/338) is the only one we still need to watch out for: In Serval since we're basically using the source ref in the target versification for pretranslations, we need to sort the pretranslations using `compareSegments=true` and then pass `compareSegments=true` to the handler when `template=source` (i.e., should be the same refs as the USFM), correct? For `template=target`, we'll want to ignore segments. This makes we wonder if we sort the rows ourselves in the handler to maintain consistency between how the refs are being compared. As long as our sorting method is a stable sorting method, this shouldn't mess up relaxed non-verse refs, right? 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341#-:-Oa611ag9vESUwqvexdF:b-p103qv)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 31 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa60cINEtLoMo4NFlSJ:-Oa60cINEtLoMo4NFlSK:b-fjksxv) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L31)):*
> ```Smalltalk
>             _curTextType.Count == 0 ? ScriptureTextType.None : _curTextType.Peek();
> 
>         protected bool DuplicateVerse { get; private set; }
> ```

I wondered about this. Do you want to add something to the editor config to flag protected fields? I think we only have something set up for protected *readonly* fields. 
___
*[`src/SIL.Machine/Corpora/VerseRefComparer.cs` line 9 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa61gNWD_tBqT9CGvT3:-Oa61gNWD_tBqT9CGvT4:b-84tga9) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/src/SIL.Machine/Corpora/VerseRefComparer.cs#L9)):*
> ```Smalltalk
> {
>     public class VerseRefComparer : IComparer<VerseRef>, IEqualityComparer<VerseRef>
> ```

If you want to keep the `ScriptureRefComparer`, should I move it to a separate file or is it appropriate to keep it with `ScriptureRef`?
___
*[`tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs` line 431 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa61R_e-27ny2AIL2MZ:-Oa61R_e-27ny2AIL2M_:b-cihgtt) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs#L431)):*
> ```Smalltalk
>         var rows = new List<UpdateUsfmRow>
>         {
>             new UpdateUsfmRow(ScrRef("MAT 2:2"), "Verse 2."),
> ```

Is this correct, @ddaspit? Is this something that could even actually happen?

Enkidu93 (commented):
In Serval, we sort the pretranslations: https://github.com/sillsdev/serval/blob/200925cd6722784fc6beb20428338267d707243a/src/Serval/src/Serval.Translation/Services/PretranslationService.cs#L148. What I'm saying is that, if sorting them is necessary, we need to make sure that the comparer we use when sorting is the same as the comparer used when advancing the rows, so maybe we should just sort the rows in the update handler itself to avoid errors caused by inconsistent sorting (i.e., the client sorts it one way, but doesn't pass the corresponding value of `compareSegments`). This is the issue I'm thinking about: https://github.com/sillsdev/machine/issues/338. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341#-:-OaLf2ZQ7b4lsoUiPqWm:b-nwg6zk)* status: all files reviewed, 2 unresolved discussions (waiting on @Enkidu93)

Enkidu93 (commented):
Done. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341#-:-OaMFMY1D9ohPp9eZAX_:bs7wumx)* status: 5 of 8 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs` line 31 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa60cINEtLoMo4NFlSJ:-OaMC0kO1Idc5py8pvP6:b32w0x) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs#L31)):*

I don't know if we can set up an editorconfig rule to catch this. Any fields that are accessible outside of a class should be accessed through a property.
</blockquote></details>

OK
___
*[`src/SIL.Machine/Corpora/VerseRefComparer.cs` line 9 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa61gNWD_tBqT9CGvT3:-OaMBzaMB1zPwu-Mo7Cf:bbr3dvq) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/src/SIL.Machine/Corpora/VerseRefComparer.cs#L9)):*

I think it makes sense to move it to a separate file.
</blockquote></details>

Sounds good. Done.
___
*[`tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs` line 431 at r3](https://reviewable.io/reviews/sillsdev/machine/341#-Oa61R_e-27ny2AIL2MZ:-OaMBy_SBY_SKn7Khxdl:b32w0x) ([raw file](https://github.com/sillsdev/machine/blob/62bf1aacf3bb8716476e5c2c1084d2f7ae775942/tests/SIL.Machine.Tests/Corpora/UpdateUsfmParserHandlerTests.cs#L431)):*

I'm not sure how realistic it is, but it is the correct behavior in this case.
</blockquote></details>

OK

ddaspit (changes requested):
From my research, it looks like sorting by `_id` would be sufficient. The main issue where the ObjectId sort order and the insertion order might not be exactly the same is when multiple processes insert docs. This is not the case for Serval. All pretranslations for a build are inserted by the same service (Serval API). We would need to add sorting support to SIL.DataAccess. In the meantime, sorting by `ScriptureRef` will work when the source is used as the template, but will not work when the target is used as the template.

> @ddaspit reviewed 4 of 5 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341#-:-OaVvfa60BcWxIpBe3wN:b-a0mh75)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 420 at r6](https://reviewable.io/reviews/sillsdev/machine/341#-OaVvaaRCvnJi7RopaTu:-OaVvaaRCvnJi7RopaTv:b-dqzgjh) ([raw file](https://github.com/sillsdev/machine/blob/166296856a0526856790ff7a663d5754364cf0a8/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L420)):*
> ```Smalltalk
>             Dictionary<string, object> rowMetadata = null;
>             int sourceIndex = 0;
>             // search the sorted rows with updated text, starting from where we left off last.
> ```

This code does assume that the rows are sorted.

ddaspit (commented):
My previous statement is not accurate.  The current implementation in Serval should work correctly for both source and target templates. We only need to sort by `ScriptureRef` prior to converting to relaxed refs for `template=Target` to work properly, which is what we are doing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/341#-:-OaVyCY9ESGS_2kCO_Kx:bmeiynj)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Issue #332, opened 2025-08-29
Custom UsfmUpdateBlock exception


Enkidu93:
As we expand the number `IUsfmUpdateBlockHandler`s it may be helpful to create a custom exception to capture update-block-related information or otherwise include it in the current exceptions if possible (since we've had a practice so far of avoiding creating deeper exception nesting). Or maybe just a `UsfmParsingException`? I continue to think that moving the exception-throwing deeper one layer might be more helpful for debugging - we may not have the project name, but we could provide verse text, ref, etc. which may be more helpful.

PR #342, opened 2025-09-25
Add update block error handling

Enkidu93:
Fixes https://github.com/sillsdev/serval/issues/764 and fixes https://github.com/sillsdev/machine/issues/332
* Add custom exception
* Add error handler
* Fix mis-ported lines

As I began coding, I realized that the idea of 'error handler' wasn't concrete enough - there are a number of places and ways we could inject an error handler. We could also design this exception a couple different ways. This is all to say, please say 'That's not at all what I was thinking', @ddaspit, if that's the case :laughing:. I figured it was simpler just to put in the PR and discuss here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/342)

### DIFF
--- a/src/SIL.Machine/Corpora/IUsfmUpdateBlockHandler.cs
+++ b/src/SIL.Machine/Corpora/IUsfmUpdateBlockHandler.cs
@@ -1,0 +1,26 @@
+using System;
+
+namespace SIL.Machine.Corpora
+{
+    public interface IUsfmUpdateBlockHandler
+    {
+        UsfmUpdateBlock ProcessBlock(UsfmUpdateBlock block);
+    }
+
+    public class UsfmUpdateBlockHandlerException : Exception
+    {
+        public UsfmUpdateBlock Block { get; }
+
+        public UsfmUpdateBlockHandlerException(string message, UsfmUpdateBlock block)
+            : base(message)
+        {
+            Block = block;
+        }
+
+        public UsfmUpdateBlockHandlerException(string message, Exception exception, UsfmUpdateBlock block)
+            : base(message, exception)
+        {
+            Block = block;
+        }
+    }
+}
--- a/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs
+++ b/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs
@@ -34,95 +34,132 @@ namespace SIL.Machine.Corpora
 
     public class PlaceMarkersUsfmUpdateBlockHandler : IUsfmUpdateBlockHandler
     {
+        private readonly Func<UsfmUpdateBlockHandlerException, bool> _errorHandler;
+
+        /// <summary>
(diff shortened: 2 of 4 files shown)


ddaspit (commented):
> @ddaspit reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/342)* status: 2 of 3 files reviewed, 1 unresolved discussion

___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 42 at r1](https://reviewable.io/reviews/sillsdev/machine/342#-Oa5DtSAAZKHxVOXc_re:-Oa5DtSAAZKHxVOXc_rf:b-60q8l0) ([raw file](https://github.com/sillsdev/machine/blob/44e775b865365f342ae642d420243709444d6dec/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L42)):*
> ```Smalltalk
>         /// <param name="errorHandler">Error handler should return true if parsing should continue; error handler should return false if exception should be thrown. Default behavior is to rethrow. </param>
>         /// </summary>
>         public PlaceMarkersUsfmUpdateBlockHandler(Func<UsfmUpdateBlockHandlerException, bool> errorHandler = null)
> ```

It would be more generic if we pass in the error handler to the `UpdateUsfmParserHandler` class and wrapped the `ProcessBlock` call in a try/catch. What do you think?

Enkidu93 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/342)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs` line 42 at r1](https://reviewable.io/reviews/sillsdev/machine/342#-Oa5DtSAAZKHxVOXc_re:-Oa5t_IB1ox18HbwnMXK:b-ym04tn) ([raw file](https://github.com/sillsdev/machine/blob/44e775b865365f342ae642d420243709444d6dec/src/SIL.Machine/Corpora/PlaceMarkersUsfmUpdateBlockHandler.cs#L42)):*

It would be more generic if we pass in the error handler to the `UpdateUsfmParserHandler` class and wrapped the `ProcessBlock` call in a try/catch. What do you think?
</blockquote></details>

Yes, it would. That was the other option that was at the forefront of my mind. 

The benefit of passing the error handler to the specific update block handlers is that it would allow more fine-grained control. For example, there may be situations in which, for a particular handler, we want the code to crash if a block cannot be processed (e.g., if we need that update to go through before being processed by a subsequent block or if it's integral to producing the USFM). If `QuotationMarkDenormalizationFirstPass` were implemented as a block handler instead, we might have that sort of issue. 

Since this isn't something we need right now, I'll update the code to just pass it to the `UpdateUsfmParserHandler`. Need be, we could move it to the individual handlers or even just pass the handler instance into the error function to provide the caller more context.

PR #344, opened 2025-09-29
Correctly parse and place verse text in verse 0

pmachapman:
Fixes https://github.com/sillsdev/serval/issues/749

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/344)

### DIFF
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -1,6 +1,5 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
-using SIL.Extensions;
 using SIL.Scripture;
 
 namespace SIL.Machine.Corpora
@@ -67,7 +66,11 @@ namespace SIL.Machine.Corpora
             string pubNumber
         )
         {
-            if (state.VerseRef.Equals(_curVerseRef) && !DuplicateVerse)
+            if (state.ChapterHasVerseZero && state.VerseRef.VerseNum == 0)
+            {
+                // Fall through for the special case of verse 0 being specified in the USFM
+            }
+            else if (state.VerseRef.Equals(_curVerseRef) && !DuplicateVerse)
             {
                 if (state.VerseRef.VerseNum > 0)
                 {
@@ -75,6 +78,8 @@ namespace SIL.Machine.Corpora
                     // ignore duplicate verses
                     DuplicateVerse = true;
                 }
+
+                return;
             }
             else if (VerseRef.AreOverlappingVersesRanges(verse1: number, verse2: _curVerseRef.Verse))
             {
@@ -82,16 +87,15 @@ namespace SIL.Machine.Corpora
                 VerseRef verseRef = _curVerseRef.Clone();
                 verseRef.Verse = CorporaUtils.MergeVerseRanges(number, _curVerseRef.Verse);
                 UpdateVerseRef(verseRef, marker);
+                return;
             }
+
+            if (CurrentTextType == ScriptureTextType.NonVerse)
+                EndNonVerseText(state);
             else
-            {
-                if (CurrentTextType == ScriptureTextType.NonVerse)
-                    EndNonVerseText(state);
-                else
-                    EndVerseText(state);
-                UpdateVerseRef(state.VerseRef, marker);
-                StartVerseText(state);
-            }
+                EndVerseText(state);
+            UpdateVerseRef(state.VerseRef, marker);
+            StartVerseText(state);
         }
 
         public override void StartPara(
@@ -259,9 +263,9 @@ namespace SIL.Machine.Corpora
 
         private void EndVerseText(UsfmParserState state)
         {
-            if (!DuplicateVerse && _curVerseRef.VerseNum > 0)
+            if (!DuplicateVerse && (_curVerseRef.VerseNum > 0 || state.ChapterHasVerseZero))
                 EndVerseText(state, CreateVerseRefs());
-            if (_curVerseRef.VerseNum > 0)
+            if (_curVerseRef.VerseNum > 0 || state.ChapterHasVerseZero)
                 _curTextType.Pop();
         }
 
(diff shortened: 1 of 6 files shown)


pmachapman:
> I'm taking a look at this Peter and stewing on whether there's a cleaner way to do it. Also, confirming that this plays nicely with the updates being made in the getting-rows-by-ref PR.

@Enkidu93 I totally understand. Probably the cleanest way I can think of would involve a custom `record` replacing `VerseRef`, as `VerseRef` doesn't like verse numbers les than 0 (technically it can have -1 when initialized, but that was quite hacky). The sorting of the rows in `ParallelTextCorpus` also added to some of the kludge - verse 0 is the only verse where the verse number appears mid-verse.

Enkidu93 (approved):
Yeah, it's a little awkward, but all the alternatives I could think of felt hacky. I'll leave this to Damien and see if he has any alternative ideas. Every time we think the parser is stable, there's another corner case :sweat_smile:. 

> The sorting of the rows in `ParallelTextCorpus` also added to some of the kludge - verse 0 is the only verse where the verse number appears mid-verse.
Could you give me more details? I tried running some `CompareTo()`s with verse-0 refs and didn't notice anything odd, but if they aren't being compared properly, that's definitely a bug.

> @Enkidu93 reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344#-:-OaVADaN0kzBEUJ8gEId:bekedfi)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 423 at r1](https://reviewable.io/reviews/sillsdev/machine/344#-OaVCuYS-B1fcgu75QYL:-OaVCuYS-B1fcgu75QYM:bkzkjza) ([raw file](https://github.com/sillsdev/machine/blob/d6a8a3f0959f460f440478a17681a1687e76ade3/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L423)):*
> ```Smalltalk
>             while (_rowIndex < _rows.Count && sourceIndex < segScrRefs.Count)
>             {
>                 // Handle the special case of verse 0, which although first in the rows, will be retrieved after other segments
> ```

After we merge in the the other PR, this will not be the case, so the extra code in this function shouldn't be needed I don't think.

Does the test you wrote pass if this is removed (since the rows are in USFM-, not sort-order)?

ddaspit (commented):
I wonder if we could always treat verse rows differently than non-verse rows, so that they can match at any point in the USFM. Let me play around with it a bit.

> @ddaspit reviewed 1 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344#-:-OaW2YCh14tNmRsCBTwY:bxyi154)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 423 at r1](https://reviewable.io/reviews/sillsdev/machine/344#-OaVCuYS-B1fcgu75QYL:-OaWJ7kzCWdS7OMlpGJ9:bathwt2) ([raw file](https://github.com/sillsdev/machine/blob/d6a8a3f0959f460f440478a17681a1687e76ade3/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L423)):*

After we merge in the the other PR, this will not be the case, so the extra code in this function shouldn't be needed I don't think.

Does the test you wrote pass if this is removed (since the rows are in USFM-, not sort-order)?
</blockquote></details>

`PlaceMarkersUsfmUpdateBlockHandlerTests.UpdateUsfm_SupportVerseZero()` fails if this logic is removed and this PR is rebased onto the `get_update_rows_by_ref` branch, but does so in a way that does not even pass through verse 0 to this function. I'll need to look into this after that PR is merged.

The `PlaceMarkersUsfmUpdateBlockHandlerTests.UpdateUsfm_SupportVerseZero()` test matches the behavior I saw when I was manually running `UsfmManualTests.ParseParallelCorpusAsync()` on a corpus that contained the problematic USFM.

pmachapman:
NB: Allow any verse to be out of order, rather than just verse 0 in the update handler.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 423 at r1](https://reviewable.io/reviews/sillsdev/machine/344#-OaVCuYS-B1fcgu75QYL:-OaX-RDr4g-uzESbepdr:bfvfehe) ([raw file](https://github.com/sillsdev/machine/blob/d6a8a3f0959f460f440478a17681a1687e76ade3/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L423)):*

`PlaceMarkersUsfmUpdateBlockHandlerTests.UpdateUsfm_SupportVerseZero()` fails if this logic is removed and this PR is rebased onto the `get_update_rows_by_ref` branch, but does so in a way that does not even pass through verse 0 to this function. I'll need to look into this after that PR is merged.

The `PlaceMarkersUsfmUpdateBlockHandlerTests.UpdateUsfm_SupportVerseZero()` test matches the behavior I saw when I was manually running `UsfmManualTests.ParseParallelCorpusAsync()` on a corpus that contained the problematic USFM.
</blockquote></details>

@Enkidu93 Your PR meant I could make my change a lot neater here. Thank you!

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 2 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344#-:-Oa_cQXvB0uyuogO4Nqe:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

Enkidu93 (approved):
> @Enkidu93 reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/344)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs` line 423 at r1](https://reviewable.io/reviews/sillsdev/machine/344#-OaVCuYS-B1fcgu75QYL:-Oa_oWtYC-T9O4mjOxap:b-iv440b) ([raw file](https://github.com/sillsdev/machine/blob/d6a8a3f0959f460f440478a17681a1687e76ade3/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs#L423)):*

@Enkidu93 Your PR meant I could make my change a lot neater here. Thank you!
</blockquote></details>

Great! More Damien's PR than mine in the end haha.

PR #357, opened 2025-11-13
Add TryParse to ScriptureRef

Enkidu93:
Needed for https://github.com/sillsdev/serval/pull/827

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/357)

### DIFF
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -45,16 +45,7 @@ namespace SIL.Machine.Corpora
 
         public static bool TryParse(string str, out ScriptureRef scriptureRef)
         {
-            try
-            {
-                scriptureRef = Parse(str);
-                return true;
-            }
-            catch (VerseRefException)
-            {
-                scriptureRef = Empty;
-                return false;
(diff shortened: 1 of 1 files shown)


ddaspit (changes requested):
> @ddaspit reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/357)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`src/SIL.Machine/Corpora/ScriptureRef.cs` line 53 at r1](https://reviewable.io/reviews/sillsdev/machine/357#-OdzB7dkEynm1Y1kAVKm:-OdzB7dkEynm1Y1kAVKn:boq18sy) ([raw file](https://github.com/sillsdev/machine/blob/918014c03c22357e2f173854ddfecadbecd019ec/src/SIL.Machine/Corpora/ScriptureRef.cs#L53)):*
> ```Smalltalk
>                 return true;
>             }
>             catch
> ```

Is there a specific exception you can catch? Generally, it is not a good idea to use a bare catch statement, since it can mask unexpected errors.

Issue #366, opened 2025-12-11
If custom.vrs filename is capitalized ("Custom.vrs"), it is not read


Enkidu93:
We ran into this recently on a published project. Is there a reason we shouldn't just check for 'Custom.vrs'? Is this common or isolated? Would this be classified as an error on their part?

PR #371, opened 2026-01-13
Check for Paratext project files case-insensitively

Enkidu93:
Fixes https://github.com/sillsdev/machine/issues/366.

I was hoping there was a more straightforward way to do this through the C# libraries, but I didn't see anything. Let me know if I've just missed it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/371)


Enkidu93:
*[`src/SIL.Machine/Corpora/FileParatextProjectFileHandler.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/machine/371#-Oj2YRoUBVRt0NGG66MT:-Oj2cWh_0gGUDJonNimI:brx4w0k) ([raw file](https://github.com/sillsdev/machine/blob/8b28605a2468ae4dd363bd97b13aef1af8d05794/src/SIL.Machine/Corpora/FileParatextProjectFileHandler.cs#L18)):*

I wonder if the overload of this method that takes a pattern and options would be better. Would `Directory.EnumerateFiles(_projectDir, fileName, new EnumerationOptions { MatchCasing = false }).Any()` work?
</blockquote></details>

I tried that and it seemed like that overload does not exist on my machine. Maybe something is wrong with my .NET set-up/version? Does it work on your machine?

ddaspit (approved):
:lgtm:

> @ddaspit made 2 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/371#-:-Oj2dISh7iBYdTIOkJfx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93).

___
*[`src/SIL.Machine/Corpora/FileParatextProjectFileHandler.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/machine/371#-Oj2YRoUBVRt0NGG66MT:-Oj2dCps9vohLsdXoM-L:b-lbtqna) ([raw file](https://github.com/sillsdev/machine/blob/8b28605a2468ae4dd363bd97b13aef1af8d05794/src/SIL.Machine/Corpora/FileParatextProjectFileHandler.cs#L18)):*

I tried that and it seemed like that overload does not exist on my machine. Maybe something is wrong with my .NET set-up/version? Does it work on your machine?
</blockquote></details>

Oh, never mind. It is only available in .NET Standard 2.1. We are using 2.0.

Issue #372, opened 2026-01-18
The USFM parser should trap the Trace warnings from VerseRef


pmachapman:
These are warnings such as:

- "Just failed to parse a verse number: Sara"
- "Just failed to parse a chapter number: 5."

Which are emitted via `Trace.TraceWarning`.

They should be trapped then emitted via the warnings in the Serval build's execution data.

ddaspit:
We need to check if this provides warnings that we cannot get from the validation status enum.

PR #381, opened 2026-02-04
Add versification warnings for invalid chapter or verse numbers in USFM

pmachapman:
Fixes: #372

Example USFM:

```
\id MAN
\c 1.
\v v1 Verse 1
```

Example warnings:

```
warnings: [
  "USFM versification error in project TEA, expected verse “”, actual verse “MAN 1.”, mismatch type InvalidChapterNumber (parallel corpus 6983d7b5415e5f0ee3585264, monolingual corpus 6983d7b5415e5f0ee3585260)",
  "USFM versification error in project TEA, expected verse “”, actual verse “MAN -1:v1”, mismatch type InvalidVerseNumber (parallel corpus 6983d7b5415e5f0ee3585264, monolingual corpus 6983d7b5415e5f0ee3585260)",
]
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/381)

### DIFF
--- a/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs
+++ b/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs
@@ -12,7 +12,9 @@ namespace SIL.Machine.Corpora
         ExtraVerse,
         InvalidVerseRange,
         MissingVerseSegment,
-        ExtraVerseSegment
+        ExtraVerseSegment,
+        InvalidChapterNumber,
+        InvalidVerseNumber
     }
 
     public class UsfmVersificationError
@@ -22,6 +24,7 @@ namespace SIL.Machine.Corpora
         private readonly int _expectedVerse;
         private readonly int _actualChapter;
         private readonly int _actualVerse;
+        private readonly string _actualValue;
         private VerseRef? _verseRef = null;
 
         public UsfmVersificationError(
@@ -41,6 +44,21 @@ namespace SIL.Machine.Corpora
             _actualVerse = actualVerse;
             _verseRef = verseRef;
             ProjectName = projectName;
+        }
+
+        public UsfmVersificationError(
+            int bookNum,
+            int expectedChapter,
+            string actualValue,
+            string projectName,
+            UsfmVersificationErrorType type
+        )
+        {
+            _bookNum = bookNum;
+            _expectedChapter = expectedChapter;
+            _actualValue = actualValue;
+            ProjectName = projectName;
+            Type = type;
         }
 
         public string ProjectName { get; private set; }
@@ -104,8 +122,14 @@ namespace SIL.Machine.Corpora
         {
             get
             {
-                if (Type == UsfmVersificationErrorType.ExtraVerse)
+                if (
+                    Type == UsfmVersificationErrorType.ExtraVerse
+                    || Type == UsfmVersificationErrorType.InvalidChapterNumber
+                    || Type == UsfmVersificationErrorType.InvalidVerseNumber
+                )
+                {
                     return "";
+                }
 
                 // We do not want to throw an exception here, and the VerseRef constructor can throw
                 // an exception with certain invalid verse data; use TryParse instead.
@@ -154,11 +178,20 @@ namespace SIL.Machine.Corpora
                 return defaultVerseRef.ToString();
             }
         }
+
         public string ActualVerseRef
         {
             get
             {
-                if (_verseRef != null)
+                if (Type == UsfmVersificationErrorType.InvalidChapterNumber)
+                {
+                    return $"{Canon.BookNumberToId(_bookNum)} {_actualValue}";
+                }
+                else if (Type == UsfmVersificationErrorType.InvalidVerseNumber)
+                {
+                    return $"{Canon.BookNumberToId(_bookNum)} {_expectedChapter}:{_actualValue}";
+                }
+                else if (_verseRef != null)
                 {
                     return _verseRef.ToString();
                 }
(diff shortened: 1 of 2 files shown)


Enkidu93 (changes requested):
> @Enkidu93 reviewed 2 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/381)* status: all files reviewed, 5 unresolved discussions (waiting on @pmachapman).

___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 49 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1l-Gc4H_00GEl-9QO:-Ol1l-Gc4H_00GEl-9QP:bjsnizr) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L49)):*
> ```Smalltalk
>         }
> 
>         public UsfmVersificationError(
> ```

Rather than adding another constructor, is there way to just check the error type in `CheckError()` itself?
___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 94 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1jiOT4JibnUa2G0YV:-Ol1jiOT4JibnUa2G0YW:b-ndsqd3) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L94)):*
> ```Smalltalk
>                     return true;
>                 }
>                 if (!_verseRef.Value.Valid)
> ```

I'm surprised that these kinds of errors (especially the bad verse numbers) aren't captured in the verse ref's valid status somehow :thinking:. It is what it is, but it'd certainly be nice if it were haha. 
___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 128 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1kD8l9NZJsta4uGBK:-Ol1kD8l9NZJsta4uGBL:b-ldg16c) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L128)):*
> ```Smalltalk
>                     Type == UsfmVersificationErrorType.ExtraVerse
>                     || Type == UsfmVersificationErrorType.InvalidChapterNumber
>                     || Type == UsfmVersificationErrorType.InvalidVerseNumber
> ```

I wonder if we could supply an expected ref. We should have some idea of what verse or chapter is expected next while parsing. If we can make a reasonable guess, I think we should add it for consistency with our attempts for other types of errors, but if we can't no worries. If the bad chapter number is something like `\c 1^`, that's pretty easy to find without an expected ref, but what about something like `\c text`: Is that possible? That would not be quite so easy to track down. 
___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 292 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1lOExEEoVVXjoC313:-Ol1lOExEEoVVXjoC314:b-7u5v4r) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L292)):*
> ```Smalltalk
> 
>             // See whether the chapter number is invalid
>             VerseRef verseRef = state.VerseRef.Clone();
> ```

Why do we need to do this? Does the `number` get changed when the state is updated so we need the raw value itself? It would be nice if we could just pass the bad verse ref to the existing constructor and check for an error in `CheckError()` - same with invalid verses. If we have to do it, we have to do it, but it'd be nice to have the error checking all in one place. 
___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 332 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1v3pz97TlGw_NujSR:-Ol1v3pz97TlGw_NujSS:b5pt2jp) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L332)):*
> ```Smalltalk
>                 {
>                     _errors.Add(versificationError);
>                     verseInError = true;
> ```

What's the particular duplicate error situation you're trying to avoid?

pmachapman (commented):
> @pmachapman made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/381)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman).

___
*[`src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs` line 49 at r1](https://reviewable.io/reviews/sillsdev/machine/381#-Ol1l-Gc4H_00GEl-9QO:-OlDRuYLCpaQAkdWwiGh:b64zuna) ([raw file](https://github.com/sillsdev/machine/blob/5ca0162a68a96fdf386fc98f71c6c299340aa9b9/src/SIL.Machine/Corpora/UsfmVersificationErrorDetector.cs#L49)):*
> If there's a way to keep all the error-checking in one place, that may be clearer and more maintainable, but it is fine as-is.

I spent the past hour trying to move the logic all in `CheckErrors()` but it got more and more convoluted with slight change breaking more and more tests :-/ Unfortunately, it looks like the logic will stay where it is.

I can only console myself with the fact is that the logic is in the same *source file*.

PR #382, opened 2026-02-10
Port key terms updates

Enkidu93:
Port https://github.com/sillsdev/machine.py/pull/257

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/382)

### DIFF
--- a/src/SIL.Machine/Corpora/ParatextProjectTermsParserBase.cs
+++ b/src/SIL.Machine/Corpora/ParatextProjectTermsParserBase.cs
@@ -117,13 +117,14 @@ namespace SIL.Machine.Corpora
                     .Descendants()
                     .Where(n => n.Name.LocalName == "TermRendering")
                     .Where(ele => ((string)ele.Attribute("Guess") ?? "false") == "false")
-                    .Select(ele => (ele.Attribute("Id").Value, ele))
+                    .Select(ele => ((string)ele.Attribute("Id"), ele))
+                    .Where(kvp => kvp.Item1 != null)
                     .Where(kvp => IsInCategory(kvp.Item1, termCategories, termIdToCategory))
                     .Where(kvp => IsInChapters(kvp.Item1, chapters, termIdToReferences))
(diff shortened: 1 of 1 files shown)


ddaspit (commented):
> @ddaspit reviewed 7 files and all commit messages, made 2 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/382)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93).

___
*[`src/SIL.Machine/Corpora/KeyTerm.cs` line 17 at r1](https://reviewable.io/reviews/sillsdev/machine/382#-Ol6vRbX-VoAWbXFAMXD:-Ol8EGCL4ejiiF8dz32V:b-p9i3tm) ([raw file](https://github.com/sillsdev/machine/blob/19e1714a4ef2c4bd1545fb45db25597587b6d76d/src/SIL.Machine/Corpora/KeyTerm.cs#L17)):*

Makes sense. Is that what you mean? `ToArray()` should shallow copy, right?
</blockquote></details>

Yes, what you have is perfect.
___
*[`src/SIL.Machine/Corpora/ParatextProjectTermsParserBase.cs` line 119 at r1](https://reviewable.io/reviews/sillsdev/machine/382#-Ol6yX_m7e_ZeG_zKZy1:-Ol8EXhDASxMoRV4Jtx7:bi1x6gy) ([raw file](https://github.com/sillsdev/machine/blob/19e1714a4ef2c4bd1545fb45db25597587b6d76d/src/SIL.Machine/Corpora/ParatextProjectTermsParserBase.cs#L119)):*

I changed it here, but didn't see any other instances quite like this. Was there another spot you had in mind? Are you saying more generally that I should cast and then use `??` with a default rather than accessing `.Value`?
</blockquote></details>

For values that might not exist, [casting is generally preferred over the `Value` property](https://learn.microsoft.com/en-us/dotnet/standard/linq/retrieve-value-element#element-might-not-exist-example).

PR #415, opened 2026-05-11
Support `VerseRef`s with segments across versifications

Enkidu93:
Includes: 
* Add test to cover alignment of verses across versifications with segments (e.g., a,b,c...). 
* Strip segments when converting/comparing refs with verse segments across versifications because `VerseRef` does not support it properly

This fixes an issue Michael recently brought up on slack: https://sil-language-software.slack.com/archives/C06A8JFBU5S/p1778328868601589. (I can create a retroactive issue if you would like). I decided to prioritize this because depending on the problem, it could be causing catastrophic misalignment - especially for silnlp - and it was in this case. 

The root of this problem is that `ChangeVersification()` (and by extension `VerseRef.CompareTo()`) exhibit very odd behavior when working with verse segments. Basically, 
```
var vrS = new VerseRef("NUM 17:1a", ScrVers.Original);
var vr = new VerseRef("NUM 17:1", ScrVers.Original);

vrS.ChangeVersification(ScrVers.English);
vr.ChangeVersification(ScrVers.English);

// Now
// vr == new VerseRef("NUM 16:36", ScrVers.English) :heavy_check_mark: 
// but 
// vrS == new VerseRef("NUM 17:1a", ScrVers.English) :exploding_head: 
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/415)

### DIFF
--- a/src/SIL.Machine/Corpora/CorporaExtensions.cs
+++ b/src/SIL.Machine/Corpora/CorporaExtensions.cs
@@ -289,10 +289,7 @@ namespace SIL.Machine.Corpora
                     continue;
 
                 VerseRef vref = scriptureRef.VerseRef;
-                if (
-                    curRef.HasValue
-                    && vref.CompareTo(curRef.Value, null, compareAllVerses: true, compareSegments: false) != 0
-                )
+                if (curRef.HasValue && VerseRefComparer.IgnoreSegments.Compare(vref, (VerseRef)curRef) != 0)
                 {
                     yield return (curTrgLineRange ? "<range>" : curTrgLine.ToString(), curRef.Value, curTrgRef.Value);
                     curTrgLineRange = curTrgLineRange || curTrgLine.Length > 0;
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -87,9 +87,7 @@ namespace SIL.Machine.Corpora
 
         public ScriptureRef ChangeVersification(ScrVers versification)
         {
-            VerseRef vr = VerseRef.Clone();
-            vr = vr.ChangeVersificationWithSegments(versification);
-            return new ScriptureRef(vr, Path);
+            return new ScriptureRef(VerseRef.ChangeVersificationWithSegments(versification), Path);
         }
 
         int IComparable<ScriptureRef>.CompareTo(ScriptureRef other)
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -88,12 +88,7 @@ namespace SIL.Machine.Corpora
         public ScriptureRef ChangeVersification(ScrVers versification)
         {
             VerseRef vr = VerseRef.Clone();
-            if (VerseRef.Versification != versification)
-            {
-                //Simplify the verse since ChangeVersification does not support verse segments
-                vr = vr.RemoveSegments();
-                vr.ChangeVersification(versification);
-            }
+            vr = vr.ChangeVersificationWithSegments(versification);
             return new ScriptureRef(vr, Path);
         }
 
--- a/src/SIL.Machine/Corpora/ScriptureRef.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRef.cs
@@ -88,7 +88,12 @@ namespace SIL.Machine.Corpora
         public ScriptureRef ChangeVersification(ScrVers versification)
         {
             VerseRef vr = VerseRef.Clone();
-            vr.ChangeVersification(versification);
+            if (VerseRef.Versification != versification)
+            {
+                //Simplify the verse since ChangeVersification does not support verse segments
+                vr = vr.RemoveSegments();
+                vr.ChangeVersification(versification);
+            }
             return new ScriptureRef(vr, Path);
         }
 
--- a/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
+++ b/src/SIL.Machine/Corpora/ScriptureRefUsfmParserHandlerBase.cs
@@ -67,7 +67,7 @@ namespace SIL.Machine.Corpora
         )
         {
             // Handle non-latin numbers
-            VerseRef verseRef = _curVerseRef.Clone();
+            VerseRef verseRef = _curVerseRef;
             verseRef.TrySetVerseUnicode(number);
 
             if (state.ChapterHasVerseZero && state.VerseRef.VerseNum == 0)
@@ -91,7 +91,7 @@ namespace SIL.Machine.Corpora
             )
             {
                 // merge overlapping verse ranges in to one range
-                verseRef = _curVerseRef.Clone();
+                verseRef = _curVerseRef;
                 verseRef.TrySetVerseUnicode(CorporaUtils.MergeVerseRanges(number, _curVerseRef.Verse));
                 UpdateVerseRef(verseRef, marker);
                 return;
--- a/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs
+++ b/src/SIL.Machine/Corpora/UpdateUsfmParserHandler.cs
@@ -776,7 +776,7 @@ namespace SIL.Machine.Corpora
             // We are using a dictionary, which uses an equality comparer. As a result, we need to change the
             // source verse ref to use the row versification. If we used a SortedList, it wouldn't be necessary, but it
             // would be less efficient.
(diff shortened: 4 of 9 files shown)


ddaspit (commented):
Did you look into `SIL.Scripture` to see if segments are being handled in any specific way by the `ChangeVersification` method?

> @ddaspit reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsWZbiT2jFfu2UbFIst:b-kzywk3)* status: 0 of 5 files reviewed, all discussions resolved.

Enkidu93:
> Did you look into `SIL.Scripture` to see if segments are being handled in any specific way by the `ChangeVersification` method?
> 
> > @ddaspit reviewed all commit messages and made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsWZbiT2jFfu2UbFIst:b-kzywk3)_ status: 0 of 5 files reviewed, all discussions resolved.

I did take a look - although it's difficult sometimes to dig through all the layers in `SIL.Scripture` to see what's really going on. From what I could tell, the look-ups for mappings between verses across versifications were exact such that if a verse had a verse segment, it didn't have a mapping to the other versification even if the verse without a verse segment did. I don't think there's any other explicit handling of verses with segments. My strategy here is to just strip the segments before converting since there isn't a guaranteed way to map those to another versification anyways.

ddaspit (commented):
I don't love entirely stripping off the segments. Maybe we could add them back in after changing the versification?

> @ddaspit made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsWripC0QUw1pLQn7Ph:bmcywux)* status: 0 of 5 files reviewed, all discussions resolved.

Enkidu93:
> I don't love entirely stripping off the segments. Maybe we could add them back in after changing the versification?
> 
> > @ddaspit made 1 comment.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsWripC0QUw1pLQn7Ph:bmcywux)_ status: 0 of 5 files reviewed, all discussions resolved.

I did consider that. Is there a reliable way to do so? I don't think we're guaranteed that the segmentation of a verse in one versification is the same in another, are we? We're definitely not guaranteed that the segment naming is the same (i.e., in some versions, it's not a,b,c - I've seen other characters used, I'm pretty sure), but I'm not even sure we're guaranteed that they really refer to the same chunk of text.

ddaspit (commented):
There probably isn't a 100% reliable way to transfer the segments, but I feel like we have to do it anyway. In most cases, it would be more correct to transfer the segment, rather than strip it. Also, there are cases where the versification has mappings specifically for verse segments. We don't want to break those mappings, so we need to detect that case and use the normal change versification logic.

> @ddaspit made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsXPjp_8-N0hizO0LhK:b-etlfuy)* status: 0 of 5 files reviewed, all discussions resolved.

Enkidu93 (commented):
I believe this is working now as we discussed yesterday. Explicit segment-to-segment (where the segments differ), cross-verse mappings will still not quite work (e.g., JON 1:1a = JON 1:3b), but I don't see any examples of this in the standard .vrs files nor do I expect it's likely (or likely to cause problems). 

> @Enkidu93 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-OsbmacJ0r7MsWeoh-Mf:b-z8lr7n)* status: 0 of 7 files reviewed, all discussions resolved (waiting on ddaspit).

ddaspit (changes requested):
> @ddaspit reviewed 7 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415)* status: all files reviewed, 2 unresolved discussions (waiting on Enkidu93).

___
*[`src/SIL.Machine/Corpora/VerseRefComparer.cs` line 23 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgqzKM-GwoO2dMMA-A:-OsgqzKM-GwoO2dMMA-B:blsq8fd) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefComparer.cs#L23)):*
> ```Smalltalk
>         public int Compare(VerseRef x, VerseRef y)
>         {
>             y = y.ChangeVersificationWithSegments(x.Versification);
> ```

Should you check if the versification is different first?
___
*[`src/SIL.Machine/Corpora/VerseRefExtensions.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgqMnM3DANTCeHAS7A:-OsgqMnM3DANTCeHAS7B:b-rutmne) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefExtensions.cs#L7)):*
> ```Smalltalk
> {
>     public static class VerseRefExtensions
> ```

I like the use of extension methods for this.
___
*[`src/SIL.Machine/Corpora/VerseRefExtensions.cs` line 13 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgoMlc1-roMLdrD9DD:-OsgoMlc1-roMLdrD9DE:byix8be) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefExtensions.cs#L13)):*
> ```Smalltalk
>             if (string.IsNullOrEmpty(verseRef.Segment()))
>             {
>                 return verseRef.Clone();
> ```

I don't think the `Clone` is necessary. `VerseRef` is a `struct` which is a value type. We probably do this in other places, but we probably shouldn't. It is just unnecessary overhead.

Enkidu93 (commented):
> @Enkidu93 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on ddaspit).

___
*[`src/SIL.Machine/Corpora/VerseRefComparer.cs` line 23 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgqzKM-GwoO2dMMA-A:-OsvUxd8Eruz5F_JUoQ7:b1k0v74) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefComparer.cs#L23)):*

Should you check if the versification is different first?
</blockquote></details>

Sorry, yes, good call. It will still work, but yes. I originally moved this check into the function and then got rid of it and forgot to move it back out.
___
*[`src/SIL.Machine/Corpora/VerseRefExtensions.cs` line 7 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgqMnM3DANTCeHAS7A:-OsvV6-g3WxweHpu28cU:bsqedf8) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefExtensions.cs#L7)):*

I like the use of extension methods for this.
</blockquote></details>

Thank you :grin: 
___
*[`src/SIL.Machine/Corpora/VerseRefExtensions.cs` line 13 at r3](https://reviewable.io/reviews/sillsdev/machine/415#-OsgoMlc1-roMLdrD9DD:-OsvVBHW4UJfJCrdC81f:b-28qxa8) ([raw file](https://github.com/sillsdev/machine/blob/9304822cedb4d4013d6a01a899c7c2d9657177ab/src/SIL.Machine/Corpora/VerseRefExtensions.cs#L13)):*

I don't think the `Clone` is necessary. `VerseRef` is a `struct` which is a value type. We probably do this in other places, but we probably shouldn't. It is just unnecessary overhead.
</blockquote></details>

Good call. If you look at the comment in SIL.Scripture on the `Clone()` function, it says:
```
// Leaving this for now to reduce code changes. Isn't really nessary when VerseRef is a struct since a = b is a copy.
```
:laughing:.

I went ahead and removed it elsewhere in the code where it shouldn't be needed.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 5 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/415#-:-Ot-OawY4iKa-cBphieN:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Enkidu93).

PR #448, opened 2026-07-02
Complexity cap: bound pathological HermitCrab parses (follow-up to rustify)

johnml1135:
## Summary

Follow-up to #446 (rustify). PR #446 made the core HermitCrab engine much faster, but grammar-induced blowups remain: certain grammar constructs (unbounded/multiple-application rules with no overt exponent, unconstrained deletion, unconstrained compounding) still cause the analysis phase to generate candidates combinatorially, sometimes taking minutes to hours for a single word. This PR implements the three-layer mitigation designed in `complexity-cap.md` (Phases 0–3; Phase 4 is FieldWorks-repo follow-up, out of scope here):

- **Layer 1 — work budget** (`b3fd2b55`): `ParseContext` propagated on `Word` exactly like `CurrentTrace`. `Morpher.MaxParseSteps`/`ParseTimeout` ship on with generous defaults. Every rule `Apply()` site checks the budget; breach is a **soft-stop** (partial results + `ParseDiagnostics`, never an exception). `RerunWithDiagnostics` re-parses one word with per-rule counters to report the top offending rule(s).
- **Layer 2 — structural bounds** (`e68f0984`): `MaxRuleApplicationsPerWord` (closes the "rule A → B → A → B" loophole that a per-rule cap alone can't catch), `MaxAnalysisShapeGrowth` (prunes analysis candidates whose hypothesized underlying form grows past the surface form), and a cascade depth cap on `PermutationRuleCascade`. All default off (no behavior change for existing consumers).
- **Layer 3 — static grammar lint** (`c8a39aeb`): `GrammarAnalyzer` with 8 stable diagnostic codes (HC0001–HC0008: no-overt-exponent affix rules, unbounded `multipleApplication`, self-feeding epenthesis/deletion, unconstrained compounding, optional-iterative lexical patterns, cyclic feeding pairs). Wired into the `hc` CLI as `hc lint` and a new `hc parse --diagnose` flag. Documented in `docs/hermitcrab-grammar-performance.md`.

Plus two small follow-up commits: doc bookkeeping (`13567446`) and a top-5-words-by-step-count diagnostic in the calibration test harness (`343515b1`).

## Calibration caveats (see `complexity-cap.md` §4.1 and §10 items 7–8)

- Real-corpus calibration against `indonesian-hc.xml`/`sena-hc.xml` showed legitimate per-word cost varies by **~1000x** between grammars, which broke the original "large multiple of one grammar's ceiling" plan. Shipped defaults (`DefaultMaxParseSteps` = 50,000,000, `DefaultParseTimeout` = 30s) are instead set with headroom above the largest legitimate word observed so far across both grammars.
- **Sena calibration is based on only a ~1% sample (72/7,121 words)** — a full-corpus re-baseline is in progress (see below) to confirm no legitimate word exceeds the shipped step default.
- **`DefaultParseTimeout` = 30s will still truncate some legitimate Sena words** (one observed at 105s). This is flagged as a genuine product tradeoff needing field input, not something resolved unilaterally in this PR — feedback welcome.

## Test plan

- [x] Full HermitCrab test suite passes (82/82), both projects build with zero warnings, csharpier clean
- [x] Determinism verified (same grammar+word → identical step count, single- and multi-threaded)
- [x] Verified against real pathological patterns (no-overt-exponent affix rule, real deletion-rule grammar)
- [ ] Full Sena corpus (7,121 words) re-baseline running in background to close the ~1% sample caveat above — will report results as a PR comment

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/448)

### DIFF
--- a/complexity-cap.md
+++ b/complexity-cap.md
@@ -92,15 +92,49 @@ the exact failure mode this plan exists to fix — an unbounded parse — remain
 out-of-the-box behavior. A generous cap that never fires for legitimate grammars but
 reliably kills runaway ones is strictly better than silence.
 
-Concrete numbers are calibrated in Phase 0 against the real corpus (§7), not guessed
-here, but the target shape is: run every word in `indonesian-words.txt` (121 words) and
-`sena-words.txt` (7,121 words) against their respective grammars on the rustify engine,
-take the observed max step count / max wall-clock time across that legitimate corpus,
-and set the default to a large multiple of that ceiling (e.g. 50–100×) so it is
-effectively invisible for real grammars but still finite. `ParseTimeout` defaults
-similarly, e.g. a flat few seconds per word — generous for interactive/FLEx single-word
-parses, still bounded for "Parse All Words" batches where one stuck word must not stall
-the run indefinitely.
+**Calibration results (2026-07-02, partial — see caveat below):** running the real
+corpus against the rustify engine shows legitimate cost varies by roughly **1000x**
+between the two grammars, which broke the original "large multiple of Indonesian's
+ceiling" plan:
+
+- `indonesian-words.txt` (121 words, 2,563-line grammar): worst observed word
+  (`mengamat-amati`, a reduplicated compound) took **10,445 steps**. `hc lint` reports
+  this grammar as clean (2 `HC0006` warnings, 0 errors), and `RerunWithDiagnostics`
+  shows the cost is a **flat distribution across ~10 rules** (~6.5% each) — legitimate
+  combinatorial interaction from compounding + reduplication, not one bad rule.
+- `sena-words.txt` (7,121 words, 33,091-line grammar): also lint-clean, but far more
+  expensive per word. The worst word sampled so far, `atawirambo`, took **14,905,517
+  steps / 105.3 seconds** — a successful, legitimate parse with the same flat
+  multi-rule-interaction shape (Sena's agglutinative verb morphology stacks many
+  candidate subject/tense/object affix slots). A separate word, `ndinakupangani`, hit
+  the (now-superseded) 10-second default timeout at only 99,584 steps, i.e. a real word
+  was previously getting truncated by the shipped default.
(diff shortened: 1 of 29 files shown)


johnml1135:
## Sena full-corpus calibration update

My own attempt to re-run the full 7,121-word Sena corpus (single-threaded, Debug build, via the `Sena_Baseline_NoWordExhaustsUnlimitedBudget`/`Sena_ShippedDefaults_NeverTrip` tests) was too slow to be practical — some individual words alone took 50+ seconds, and it was aborted after ~1 hour at 283/7,121 words rather than run for what looked like 20+ hours.

A separate, much more efficient investigation (sharded 8-way, Release-mode instrumented harness — see `docs/hermitcrab-parse-algorithm-analysis.md`, currently on a sibling `parse-optimization` branch and not yet committed anywhere) got further and found:

- p90 ≈ 2,000,000 steps; ~16% of Sena words exceed 1,000,000 steps
- Worst observed so far ≥ 39,900,000 steps (`kukucitirani`) — under the shipped 50,000,000-step `DefaultMaxParseSteps`, but with less headroom than the earlier ~1% sample suggested, and not yet a confirmed corpus-wide max
- `DefaultParseTimeout` = 30s trips on **dozens** of legitimate Sena words (100–250s single-threaded), not just the one word noted originally
- **New finding:** `cinacemerwa` (37.5M steps, 0 valid parses) crashed the NUnit test host outright, apparently from memory pressure — independent of the step/timeout budgets, which bound steps and wall-clock but not allocations

Pushed as `c1d7db64`, updating `complexity-cap.md` §10 items 7–8 and adding item 9 (the OOM-crash finding). All still open — full corpus re-baseline and the timeout product decision remain unresolved, and it's now an open question whether budget #3 (allocation/candidate-count ceiling) is needed or whether the `parse-optimization` branch's in-progress algorithmic fixes (nogood cache / memoization, directly targeting the same redundant-search root cause) make it moot.

PR #449, opened 2026-07-02
Fix LT-22600: Show Hermit Crab rule statistics

jtmaxwell3:
Adds code to instrument rule applications by creating an InstrumentedRule class for the rules that should be instrumented.  Changes the code for creating rule classes to record data needed for instrumentation.  Changes the Apply functions to collect statistics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/449)


jtmaxwell3:
I no longer think that this is necessary or the best approach.

PR #452, opened 2026-07-07
Support LT-22605: Add ability to limit HermitCrab parses

jtmaxwell3:
MaxUnapplications was an experimental variable that never got incorporated into FieldWorks.  I decided to rename it MaxAlternatives for clarity and to change it to throw a TimeoutException since the MaxUnapplications implementation truncated results without telling the user and since it could produce different results with different parses of the same word because of multi-threading.

MaxAlternatives limits the number of alternatives attempted by each stratum during analysis.  This is where most of the blowup occurs since rules and templates within a stratum are unordered and since rules and templates are optional during unapplication (analysis).  The stratum checks MaxAlternatives as it iterates over the results produced by ApplyTemplates and ApplyMorphological rules.  This terminates without a full enumeration because ApplyTemplates and ApplyMorphologicalRules use lazy evaluation (yield return).  However, _mrulesRule and _mTemplatesRule do not use lazy evaluation, so we set MaxAlternatives on them as well.  Because of this, the maximum number of alternatives can be some multiple of MaxAlternatives in the worst case.  We could also pass in the current number of alternatives to _mrulesRule and _mTemplatesRule, but we probably don't need to be that precise.  We are just trying to avoid words taking forever to parse.

MaxAlternatives also limits the number of syntheses attempted.

We could also limit the number of alternatives considered within each stratum during synthesis, but the number of alternatives generated is usually proportionate to the number of syntheses attempted because rules are obligatory in the synthesis direction and there are rarely alternative orderings of the rules that match.  The current limits should be enough to bound the amount of time spent parsing a word.

NB: I had to wrap the bodies of Parallel.ForEach statements with try ... catch in ParallelCombinationRuleCascade and ParallelPermutationRuleCascade in order to get the TimeoutException to propagate correctly (noted by Devin).

I tested this against Aweti-opap-no-go from Andy Black by setting MaxAlternatives to 1000 and all of the words finished in a reasonable amount of time and the words that were limited gave a reasonable error message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/452)


johnml1135:
Hello John,

I am still trying to get a grapple on how this all works (getting better).  I like the approach - having a deterministic "kill the parse if it takes too long" and it seems like a meaningful improvement over what we have.  I investigated it with AI and here are some comments that seem reasonable to me:

```
The robustness gaps are mostly small, contained fixes:

1. Unwrap the parallel exception — either wrap the Parallel.ForEach body in ParallelCombinationRuleCascade with the same try/catch-and-rethrow pattern Morpher.Synthesize already uses (Morpher.cs:330-338), or move the check outside the parallel region. Without this, the feature is broken on the default configuration it's meant to serve.
2. Count with an Interlocked counter instead of output.Count() — fixes both the O(n²) counting on ConcurrentStack and the interleaving-dependent throw-or-not behavior near the boundary. One line of state, restores the determinism claim.
3. Push the cap into _prulesRule too — the CheckMaxAlternatives call is already sitting in LinearRuleCascade; it just never gets armed. One extra SetMaxAlternatives call.
4. Fix the mixed condition — use _maxAlternatives consistently instead of _morpher.MaxAlternatives > 0 && count >= _maxAlternatives, which removes the latent instant-throw on the un-initialized path, and align the >= vs > off-by-one while there.
5. A dedicated exception type (e.g. MaxAlternativesExceededException : Exception) rather than TimeoutException would let callers distinguish "grammar too ambiguous for this word" from actual timeouts, and carries the count for the error message. Cheap, but an API decision the author may prefer to keep as-is if FLEx already handles TimeoutException.

The one thing that isn't a small tweak is that it bounds breadth, not depth — a single slow pattern match with few alternatives sails past the cap. Fixing that properly means threading a real cancellation/deadline through the matcher, which is a much bigger change. But that's arguably fine as a scope cut: the observed 30-minute words are alternative-explosion cases, and this handles those. The design smell of mutable SetMaxAlternatives state on shared compiled rules is similar — worth noting in review, not worth blocking over, since every current caller sets a constant value.

So: sound idea, right layer, and items 1–4 are each a few lines. Item 1 is the only one I'd call a must-fix before merge, since it defeats the feature in the default build.
```

Looking forward to this landing!

jtmaxwell3:
@johnml1135:

1: I don’t understand why "the feature is broken on the default configuration".  It worked fine for me.
2: I’m not sure what is meant by “Count with an Interlocked counter”.
3: _prulesRule always has exactly one alternative because of how phonological rules work, so there is nothing to bound.
5: I don’t think we need a dedicated exception type.  There are no other timeouts, and even though it isn’t technically a timeout, it effectively is.

 I’m fine with only bounding the breadth, not the depth.  I’m just trying to prevent an exponential explosion, and I don’t think that you can have an exponential depth.

ddaspit (changes requested):
> @ddaspit reviewed 12 files and all commit messages, and made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452)* status: all files reviewed, 11 unresolved discussions (waiting on jtmaxwell3).

___
*[`src/SIL.Machine/Rules/ParallelCombinationRuleCascade.cs` line 59 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfFd7LA97eOq6yy4XV:-OxfFd7LA97eOq6yy4XW:bgan66f) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/ParallelCombinationRuleCascade.cs#L59)):*
> ```Smalltalk
>                                         output.PushRange(results);
>                                         Interlocked.Add(ref alternativeCount, results.Length);
>                                         CheckMaxAlternatives(alternativeCount);
> ```

`CheckMaxAlternatives(Interlocked.Add(ref alternativeCount, results.Length))` would be better.
___
*[`src/SIL.Machine/Rules/RuleBatch.cs` line 47 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDZT35-NHw74GXXDZ:-OxfDZT35-NHw74GXXD_:b33kfil) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleBatch.cs#L47)):*
> ```Smalltalk
>         }
> 
>         public void SetMaxAlternatives(int maxAlternatives)
> ```

This should be a property.
___
*[`src/SIL.Machine/Rules/RuleBatch.cs` line 58 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCjbn1GlisMPknc8L:-OxfCjbn1GlisMPknc8M:b-4g47pi) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleBatch.cs#L58)):*
> ```Smalltalk
>             {
>                 output.UnionWith(rule.Apply(input));
>                 if (_maxAlternatives > 0 && output.Count >= _maxAlternatives)
> ```

I believe that this should be `output.Count > _maxAlternatives`.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 52 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDawSEhs8B5Q-l4q6:-OxfDawSEhs8B5Q-l4q7:b33kfil) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L52)):*
> ```Smalltalk
>         }
> 
>         public void SetMaxAlternatives(int maxApplications)
> ```

This should be a property.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 57 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDQNF3xLluU4dWvVx:-OxfDQNF3xLluU4dWvVy:b-gnm74n) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L57)):*
> ```Smalltalk
>         }
> 
>         public void CheckMaxAlternatives(int alternativeCount)
> ```

This should be `protected`.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 59 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCpHy9nPO91gf_pkc:-OxfCpHy9nPO91gf_pkd:b-fp7ae) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L59)):*
> ```Smalltalk
>         public void CheckMaxAlternatives(int alternativeCount)
>         {
>             if (_maxAlternatives > 0 && alternativeCount >= _maxAlternatives)
> ```

I believe that this should be `alternativeCount > _maxAlternatives`.
___
*[`src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs` line 143 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCwaI9B_p6dDISZPH:-OxfCwaI9B_p6dDISZPI:b-fp7ae) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs#L143)):*
> ```Smalltalk
>             {
>                 alternativeCount++;
>                 if (_maxAlternatives > 0 && alternativeCount >= _maxAlternatives)
> ```

I believe that this should be `alternativeCount > _maxAlternatives`.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 75 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDEC581sZkF8S5HKC:-OxfDEC581sZkF8S5HKD:btfj1w3) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L75)):*
> ```Smalltalk
> 
>         /// <summary>
>         /// MaxUnapplications limits the number of unapplications to make it possible
> ```

This comment needs to be fixed.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 322 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCGRx5WtzJdmh6bS_:-OxfCGRx5WtzJdmh6bSa:b-q8zw2s) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L322)):*
> ```Smalltalk
>                                 foreach (Word alternative in synthesisWord.ExpandAlternatives())
>                                 {
>                                     alternativeCount++;
> ```

You should use `Interlocked.Increment` here to avoid a race condition.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 323 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCPnJAXpp_NjoJz1L:-OxfCPnJAXpp_NjoJz1M:b-6t573d) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L323)):*
> ```Smalltalk
>                                 {
>                                     alternativeCount++;
>                                     if (MaxAlternatives > 0 && alternativeCount >= MaxAlternatives)
> ```

I believe that this should be `alternativeCount > MaxAlternatives`.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 324 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDn0E7cIrhZvzLd-X:-OxfDn0E7cIrhZvzLd-Y:bfax0sv) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L324)):*
> ```Smalltalk
>                                     alternativeCount++;
>                                     if (MaxAlternatives > 0 && alternativeCount >= MaxAlternatives)
>                                         throw new TimeoutException("MaxAlternatives exceeded");
> ```

I would prefer this be a custom exception. `TimeoutException` has a bit of a different meaning.

jtmaxwell3 (commented):
> @jtmaxwell3 made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452)* status: 5 of 13 files reviewed, 11 unresolved discussions (waiting on ddaspit).

___
*[`src/SIL.Machine/Rules/ParallelCombinationRuleCascade.cs` line 59 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfFd7LA97eOq6yy4XV:-OxfxQ-k2s6s591pmaph:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/ParallelCombinationRuleCascade.cs#L59)):*

`CheckMaxAlternatives(Interlocked.Add(ref alternativeCount, results.Length))` would be better.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/RuleBatch.cs` line 47 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDZT35-NHw74GXXDZ:-OxfxRqN22L6KJLkCU6o:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleBatch.cs#L47)):*

This should be a property.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/RuleBatch.cs` line 58 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCjbn1GlisMPknc8L:-OxfxUGcDZ6OrwtOFrSH:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleBatch.cs#L58)):*

I believe that this should be `output.Count > _maxAlternatives`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 52 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDawSEhs8B5Q-l4q6:-OxfxVYN876wQc9RnyzS:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L52)):*

This should be a property.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 57 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDQNF3xLluU4dWvVx:-OxfxWqd9LQ87lOe1mXf:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L57)):*

This should be `protected`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/RuleCascade.cs` line 59 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCpHy9nPO91gf_pkc:-OxfxY3k7XLxntSe5zse:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine/Rules/RuleCascade.cs#L59)):*

I believe that this should be `alternativeCount > _maxAlternatives`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs` line 143 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCwaI9B_p6dDISZPH:-Oxfxb5-EiMeHS-3eN4P:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/AnalysisStratumRule.cs#L143)):*

I believe that this should be `alternativeCount > _maxAlternatives`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 75 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDEC581sZkF8S5HKC:-Oxfxd-G9qxxEB4RclJP:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L75)):*

This comment needs to be fixed.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 322 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCGRx5WtzJdmh6bS_:-OxfxeBuBfdxbdhSiRux:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L322)):*

You should use `Interlocked.Increment` here to avoid a race condition.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 323 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfCPnJAXpp_NjoJz1L:-OxfxgvO8cYDuH4uIj88:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L323)):*

I believe that this should be `alternativeCount > MaxAlternatives`.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 324 at r2](https://reviewable.io/reviews/sillsdev/machine/452#-OxfDn0E7cIrhZvzLd-X:-Oxfxi7k9JcVmfX4yYAt:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/ae9bd9a4f41000063f15e280d84f5eb8a6bbfcf4/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L324)):*

I would prefer this be a custom exception. `TimeoutException` has a bit of a different meaning.
</blockquote></details>

Done.

ddaspit (commented):
> @ddaspit reviewed 8 files and all commit messages, made 3 comments, and resolved 11 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452)* status: all files reviewed, 3 unresolved discussions (waiting on jtmaxwell3).

___
*[`src/SIL.Machine/Rules/MaxAlternativesExceeded.cs` at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgBYc4CYXT2Zub2Zfd:-OxgBYc4CYXT2Zub2Zfe:b-6ssrb) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine/Rules/MaxAlternativesExceeded.cs)):*
This file should be renamed `MaxAlternativesExceededException.cs` to match the class name.
___
*[`src/SIL.Machine/Rules/ParallelPermutationRuleCascade.cs` line 56 at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgB-3u50oLAdflA-tC:-OxgB-3u50oLAdflA-tD:b9mk7sf) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine/Rules/ParallelPermutationRuleCascade.cs#L56)):*
> ```Smalltalk
>                                 {
>                                     output.PushRange(results);
>                                     Interlocked.Add(ref alternativeCount, results.Length);
> ```

Same pattern as `ParallelCombinationRuleCascade`, `CheckMaxAlternatives(Interlocked.Add(ref alternativeCount, results.Length))` would be better.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 323 at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgBFgM8G7U92aYAPUS:-OxgBFgN7mAZ0yVlPQl9:bn8rdie) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L323)):*
> ```Smalltalk
>                                 {
>                                     Interlocked.Increment(ref alternativeCount);
>                                     if (MaxAlternatives > 0 && alternativeCount > MaxAlternatives)
> ```

Same pattern as the rule cascades, `if (MaxAlternatives > 0 && Interlocked.Increment(ref alternativeCount) > MaxAlternatives)` would be better.

jtmaxwell3 (commented):
> @jtmaxwell3 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on ddaspit).

___
*[`src/SIL.Machine/Rules/ParallelPermutationRuleCascade.cs` line 56 at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgB-3u50oLAdflA-tC:-OxgJ2S83yVicUAGGIU3:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine/Rules/ParallelPermutationRuleCascade.cs#L56)):*

Same pattern as `ParallelCombinationRuleCascade`, `CheckMaxAlternatives(Interlocked.Add(ref alternativeCount, results.Length))` would be better.
</blockquote></details>

Done.
___
*[`src/SIL.Machine.Morphology.HermitCrab/Morpher.cs` line 323 at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgBFgM8G7U92aYAPUS:-OxgJ4KH0riy0sfXXnHt:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine.Morphology.HermitCrab/Morpher.cs#L323)):*

Same pattern as the rule cascades, `if (MaxAlternatives > 0 && Interlocked.Increment(ref alternativeCount) > MaxAlternatives)` would be better.
</blockquote></details>

Done.
___
*[`src/SIL.Machine/Rules/MaxAlternativesExceeded.cs` at r3](https://reviewable.io/reviews/sillsdev/machine/452#-OxgBYc4CYXT2Zub2Zfd:-OxgIyhl1YjEVdztNvTq:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/d6a80da1ea8c2fd8f5d4f998051b45d9b0354579/src/SIL.Machine/Rules/MaxAlternativesExceeded.cs)):*

This file should be renamed `MaxAlternativesExceededException.cs` to match the class name.
</blockquote></details>

Done.

ddaspit (commented):
The build seems to be failing. It looks like a formatting error.

> @ddaspit partially reviewed 3 files and all commit messages, made 1 comment, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452#-:-OxgKFUTAY-C7h_LLj-o:b-wwfz6f)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/452#-:-Oxkekx9Cc2drJEvByow:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #453, opened 2026-07-10
LT-22613: fix analysis-length bound over-pruning for count-mismatched rewrite subrules

johnml1135:
## Summary

Fixes LT-22613: `Morpher.ParseWord`'s default (non-tracing) path returned 0 parses for words a grammar should legitimately parse, while the traced path (which happens to skip this gate) correctly returned 1. The JIRA issue's original write-up hypothesized the cause was `AnalysisScope`'s nogood-memoization cache — that was investigated and **disproven** (toggling the memo cache independently of the real cause changes nothing); see the corrective comment added to LT-22613 and the corrected analysis doc on `bug/csharp-nogood-cache-tracing-divergence` (`rust/docs/bugs/csharp-nogood-cache-tracing-divergence.md`).

**Actual root cause:** `GrammarAnalyzer.ComputeMaxAnalysisLength` (Phase 4 "Gate B" of `parse-optimization.md`) computes a grammar-wide upper bound on analysis-form length, used by `AnalysisStratumRule.Apply` to prune candidates early — but only on the default, non-tracing path. For any phonological rewrite subrule where `Lhs.Count != Rhs.Count`, the analysis-side unapplication (`EpenthesisAnalysisRewriteRuleSpec`/`NarrowAnalysisRewriteRuleSpec`) never shrinks the candidate shape — it adds new `Optional` (not `Deleted`) nodes, which `Shape.SegmentCount()` still counts. The bound assumed growth of `Lhs.Count - Rhs.Count` per match site; real growth is always `Lhs.Count`, regardless of direction. This under-counted the true bound and pruned valid candidates as "too long."

- `9a43d251` fixes the insertion/expansion case (`Rhs.Count > Lhs.Count`, epenthesis when `Lhs` is empty).
- `93e52dc9` broadens the fix to any count mismatch after an independent adversarial review found the identical bug shape in deletion/coalescence subrules (`Lhs.Count > Rhs.Count`), and removes the now-unreachable deletion-reapplication compounding logic (and the `DeletionReapplications` parameter it consumed) as dead code.

Base is `parse-optimization` (not `master`): `GrammarAnalyzer`/`AnalysisScope`/`MaxAnalysisLength` only exist on this branch (PR #451, based on `hc-rustify` PR #446).

## Test plan

- [x] `dotnet build` — full solution, 0 warnings/errors
- [x] `dotnet test tests/SIL.Machine.Morphology.HermitCrab.Tests` — 71/71 passed
- [x] New regression tests, independently verified red-then-green against each parent commit:
  - `ComputeMaxAnalysisLength_ReturnsNull_ForInsertionRewriteRule` / `ForDeletionRewriteRule` (unit-level bound assertions)
  - `EpenthesisRuleWithMinimalLexicon` (end-to-end minimal-lexicon parse, default vs. traced path)
- [x] JIRA LT-22613 updated with the corrected root cause and these commit references

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/453)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/AnalysisStateKey.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/AnalysisStateKey.cs
@@ -87,10 +87,7 @@ namespace SIL.Machine.Morphology.HermitCrab
                 return false;
             if (!_shape.ValueEquals(other._shape))
                 return false;
-            if (
-                !_syntacticFS.ValueEquals(other._syntacticFS)
-                || !_realizationalFS.ValueEquals(other._realizationalFS)
-            )
+            if (!_syntacticFS.ValueEquals(other._syntacticFS) || !_realizationalFS.ValueEquals(other._realizationalFS))
                 return false;
             return RuleCountsEqual(_ruleCounts, other._ruleCounts);
         }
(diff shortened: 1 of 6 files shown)


johnml1135:
THis was incorrectly made - this bug was in the opimazation code, not main.

PR #457, opened 2026-07-17
Stack RUSTIFY's array/COW rearchitecture on top of memoization

johnml1135:
## Summary

Stacked on #456 (memoization). Answers a question that PR left open: does the missing
copy-on-write `Shape`/`ShapeNode` (an array-backed rearchitecture from the previously-reviewed
`hc-rustify` branch, PR #446) meaningfully improve on memoization alone, or is it a "kind of, but
not really" difference?

This PR's diff is exactly the general allocation/data-structure rearchitecture from PR #446
(flat/COW `Shape`, int-offset FST traversal, cheap hash overrides, etc.) cherry-picked onto
`feature/memoization` and merge-resolved — **not** the FST inverse-chain analyzer or grammar
linter from the separate, still-experimental `fst-advisor` exploration branch. That's a
different optimization strategy (compile-to-FST vs. memoize-the-search) with its own,
independently-confirmed correctness gap on pathological words; out of scope here.

**Cost of this PR**: a much larger diff than #456 (112 files vs. 11) touching the engine's core
`Shape`/`ShapeNode` data representation, not just the analysis cascade. It was independently
reviewed once already as PR #446; this PR re-verifies it against the current codebase plus
memoization's new code paths (merge notes below), but reviewers should weigh that scope against
the numbers below, not just the numbers alone. #456's off-by-default safety property is
unchanged here — the array/COW rearchitecture activates for every `Morpher`, memoized or not,
same as it did in #446, and the memo itself is still only reachable via
`maxDegreeOfParallelism: 1`.

## Result: a real improvement, apples to apples

Same methodology as `memoization.md` (#456): canonical analysis-set signature comparison (never
byte-identical), same two heavy words used to verify memoization alone, same local uncommitted
Sena grammar.

| Word | memo-on | memo-off | Ratio | Divergences | vs. memoization alone (#456) |
|---|---|---|---|---|---|
| H1 (known heavy, positive analysis) | 9.17s | 90.0s | **9.81x** | 0 | was 22.1s/136.1s/6.2x — faster on both sides, better ratio |
| H2 (second heavy word) | 44.8s | 263.8s | **5.89x** | 0 (both empty) | was >400s both sides — never completed at all before |

H2 is the headline: the word memoization alone could not get through even a 400s budget now
completes cleanly in well under a minute. One honest caveat on H2, spelled out in
`rustify-cow.md`: it resolves to zero valid analyses on both sides in this grammar, so the
soundness check there is "both sides agree on empty," not a positive multi-analysis match like
H1 — still non-vacuous (434,628 nogood hits recorded), just a weaker confirmation.

## Cross-language aggregate and multithreading (4 threads)

The H1/H2 numbers above are a deep-dive on two specific pathological words. Separately, to get an
apples-to-apples read across grammars (not just the worst known words), the corpus-verification
harness was re-run with a 200s per-word timeout on larger slices — 250 words each for Sena and
Amharic, all 121 words that exist in the Indonesian list — across four states: **baseline**
(today's shipped parallel default), **memo** (#456 alone), **cow-only** (this PR, no memo), and
**memo+cow** (this PR stacked on #456). Zero analysis-set divergences across all 12 runs.

| Language | State | Timed out (>200s) | Mean (ms) | p50 (ms) | p95 (ms) |
|---|---|---|---|---|---|
| Sena | baseline | 2/250 | 3362.4 | 112.9 | 16444.2 |
| Sena | memo | 0/250 | 1856.0 | 106.1 | 6082.4 |
| Sena | **memo+cow** | **0/250** | **932.5** | **61.1** | **3993.1** |
| Amharic | baseline | 6/250 | 7261.9 | 297.5 | 35374.3 |
| Amharic | memo | 6/250 | 6460.9 | 299.5 | 31561.4 |
| Amharic | **memo+cow** | **1/250** | **2943.2** | **62.8** | **14714.6** |
| Indonesian | baseline | 0/121 | 17.9 | 5.8 | 52.5 |
| Indonesian | memo | 0/121 | 20.3 | 6.1 | 62.7 |
| Indonesian | memo+cow | 0/121 | 15.1 | 4.2 | 47.7 |

**Amharic is the important row**: memoization alone barely moves the timeout count (6 → 6) or
the mean (~11% better). This PR's rearchitecture is what actually fixes Amharic's pathological
words (timeouts 6 → 1, mean more than halved) — full breakdown, including the cow-only isolation
row, in `rustify-cow.md`.

**Words/second with 4x multithreading** (new: `MemoFourThreadThroughput_WordsPerSecond`, same
word slices, 4 worker threads sharing one sequential+memo `Morpher`, 200s per-word watchdog;
1-thread Sena/Amharic figures are derived (`1000/meanMs`) from the table above, not separately
measured — the Indonesian row *is* separately measured and the derived estimate there overstated
the real number by ~21%, so treat the derived figures as approximate):

| Language | Threads | memo (words/sec) | memo+cow (words/sec) | memo+cow vs memo |
|---|---|---|---|---|
| Sena | 1 (derived) | 0.54 | 1.07 | — |
| Sena | 4 | 1.11 | **2.40** | **2.16x** |
| Amharic | 1 (derived) | 0.15 | 0.34 | — |
| Amharic | 4 | 0.11 (18 timeouts) | **0.40** (4 timeouts) | 3.64x (confounded by the timeout-count gap, not a clean multiplier) |
| Indonesian | 1 (measured) | 40.75 | 56.55 | — |
| Indonesian | 4 | 46.50 | **100.35** | **2.16x** |

This confirms "COW enables better multithreading" directionally, but the honest shape of it
differs from a flat "extra 2x": on Sena and Indonesian, this PR's 1→4 thread scaling really is
better than memo-only's (2.24x vs 2.06x; 1.77x vs 1.14x, the latter pair both real measurements).
On Amharic, the finding isn't extra speed so much as **robustness under contention** — at 4
threads, memo-only starts missing the 200s watchdog on 3x more words (18 vs 6 single-threaded)
than it did running alone, while memo+cow only loses the watchdog on 1 more (4 vs 1). COW is the
difference between concurrent Amharic parsing holding up and falling over, more than it is a
uniform 2x bonus.

## Merge notes and an anomaly, not hidden

Full writeup — including the merge-resolution reasoning for the 4 files both branches touched —
is in `rustify-cow.md`. One thing worth flagging directly rather than leaving to the doc: the
first H1 measurement, at a 240s budget, timed out, which is odd since 9.17s + 90.0s is nowhere
near 240s. A re-run at 600s completed cleanly with the numbers above. The doc treats the
transient-load explanation (this session ran many consecutive heavy benchmarks right before) as
**plausible, not proven** — the memo-hit counters from the failed run matched the completed
run's final tally, which is suggestive that the computation had actually finished by read-time,
but not conclusive. Flagging this as an open, not fully closed, question rather than asserting
it's settled.

## Test plan

- [x] `dotnet build Machine.sln`
- [x] `dotnet csharpier check .`
- [x] HermitCrab test suite: 80/80 (78 from #456 + 2 new from this branch)
- [x] Full `Machine.sln` test suite green
- [x] Corpus verification harness run manually against local Sena/Amharic/Indonesian grammars — 0 divergences across the single-word deep-dive, the 250/250/121-word cross-language aggregate, and the 4-thread throughput matrix (12 + 6 runs, all states)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/457)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,13 @@ tests/SIL.Machine.Tests/Corpora/TestData/usfm/target/*
 tests/SIL.Machine.Tests/Corpora/TestData/project/*
 tests/SIL.Machine.Tests/Corpora/TestData/pretranslations.json
 .idea
+
+# Local-only HermitCrab benchmark fixtures (real Sena/Indonesian grammars + word lists, used
+# for ad hoc perf/allocation testing) + FieldWorks project backups. Large and/or not licensed
+# for this repo, so they stay untracked; any [Explicit] benchmark that wants them falls back to
+# the tracked samples/data/en-hc.xml when they're absent.
+*.fwbackup
+samples/data/sena-hc.xml
+samples/data/sena-words.txt
+samples/data/indonesian-hc.xml
+samples/data/indonesian-words.txt
--- a/rustify-cow.md
+++ b/rustify-cow.md
@@ -135,11 +135,110 @@ RUSTIFY doesn't measurably change the typical-word tradeoff in either direction 
 existing honest caveat in `memoization.md` §6 (sequential+memo loses to parallel-default on
 typical words, mostly from the lost thread) stands unchanged.
 
+## Cross-language aggregate: 250/250/121-word slices, four states, 200s watchdog
+
+The H1/H2 deep-dive above targets two specific pathological words with a long (600s) budget.
+Separately, to get a genuine apples-to-apples read across grammars (not just the worst known
+words), the same corpus-verification harness was re-run with a 200s per-word timeout on larger
+slices: 250 words each for Sena and Amharic (both have thousands of real words available; 250 is
+an arbitrary "big enough" cut), and all 121 words that exist in the Indonesian list (that corpus
+is just small — not a chosen cutoff). Four states compared: **baseline** (today's shipped
+parallel default, unmemoized), **memo** (#456 alone), **cow-only** (parallel + this PR's
+rearchitecture, no memo — free bonus from the harness), and **memo+cow** (this PR, stacked).
+
+| Language | State | Completed/N | Timed out (>200s) | Mean (ms) | p50 (ms) | p95 (ms) |
+|---|---|---|---|---|---|---|
+| Sena | baseline | 248/250 | 2 | 3362.4 | 112.9 | 16444.2 |
+| Sena | memo | 250/250 | 0 | 1856.0 | 106.1 | 6082.4 |
+| Sena | cow-only | 249/250 | 1 | 2232.4 | 72.5 | 11067.7 |
+| Sena | **memo+cow** | 250/250 | **0** | **932.5** | **61.1** | **3993.1** |
+| Amharic | baseline | 244/250 | 6 | 7261.9 | 297.5 | 35374.3 |
+| Amharic | memo | 244/250 | 6 | 6460.9 | 299.5 | 31561.4 |
+| Amharic | cow-only | 249/250 | 1 | 3208.8 | 70.4 | 15244.6 |
+| Amharic | **memo+cow** | 249/250 | **1** | **2943.2** | **62.8** | **14714.6** |
+| Indonesian | baseline | 121/121 | 0 | 17.9 | 5.8 | 52.5 |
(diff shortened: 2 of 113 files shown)


johnml1135:
We are intentioanlly not doing this because of the scope change.

PR #464, opened 2026-07-25
Fix unhandled InvalidCastException when a rewrite/metathesis rule has a top-level Quantifier

johnml1135:
## The bug

`Constraint<TData, TOffset>` (`src/SIL.Machine/Matching/Constraint.cs:12-13`) and `Quantifier<TData, TOffset>` (`src/SIL.Machine/Matching/Quantifier.cs:13-14`) are **both direct subclasses** of `PatternNode<TData, TOffset>` — siblings, with no inheritance relationship between them.

The HermitCrab rule specs assume every top-level child of a rule's target (Lhs) or replacement (Rhs) pattern — or a metathesis switch group — is a `Constraint`, and cast accordingly with `Children.Cast<Constraint<Word, ShapeNode>>()` or a direct `(Constraint<Word, ShapeNode>)` cast:

- `PhonologicalRules/SynthesisRewriteRuleSpec.cs:33` (and `:93`, lazily, via `Pattern.Acceptable`)
- `PhonologicalRules/FeatureAnalysisRewriteRuleSpec.cs:42-43`
- `PhonologicalRules/NarrowAnalysisRewriteRuleSpec.cs:45` (lazily, in `Unapply`)
- `PhonologicalRules/EpenthesisAnalysisRewriteRuleSpec.cs:18`
- `PhonologicalRules/SynthesisMetathesisRuleSpec.cs:31`
- `PhonologicalRules/AnalysisMetathesisRuleSpec.cs:53`

But the DTD's `PhoneticSequence` element (`HermitCrabInput.dtd:515`) allows an `OptionalSegmentSequence` in exactly the same positions as a plain `Segment`/`SimpleContext` — and `OptionalSegmentSequence` itself (`HermitCrabInput.dtd:560`) carries an `id` attribute explicitly documented as "only needed for a metathesis rule", so it's legal there too. `XmlLanguageLoader.LoadPatternNodes` (~1405-1415, ~1493-1505) builds a real `Quantifier` for an `OptionalSegmentSequence` with no LHS-vs-environment distinction.

So a **DTD-legal grammar** — a rewrite rule whose target or replacement contains a top-level optional/repeated segment sequence — crashes with an unhandled `InvalidCastException` during `Morpher` construction. Depending on which of the sites above runs first, it either propagates raw straight out of `new Morpher(...)` (the synthesis side has no try/catch around rule compilation), or gets caught by a pre-existing, unrelated catch-all in `AnalysisStratumRule.CompilePhonologicalRule` — but even then the message is uninformative (just names the rule; the real cause is buried in an opaque `InnerException`).

Quantifiers already work correctly in a rule's *environment* (`LeftEnvironment`/`RightEnvironment` — see the existing `QuantifierRules` test) since environment matching doesn't go through this cast. The bug is specific to the target/replacement/switch-group position.

## Reproduction

Verified directly (not hypothetical) — found while building an independent FST-based morphological analyzer/proposer ([PanGloss](https://github.com/sillsdev)) against this library; a grammar under construction hit this.

Two new tests in `RewriteRuleTests.cs` build a `RewriteRule` directly (bypassing the XML loader, matching the rest of that file) with a top-level `Quantifier` in the Lhs and, separately, the Rhs, then construct a `Morpher`:

- `QuantifierInTargetIsRejectedWithClearError` — before the fix, throws a **raw, unhandled `InvalidCastException`** straight out of `new Morpher(...)` (confirmed against unmodified `master`).
- `QuantifierInReplacementIsRejectedWithClearError` — before the fix, throws a `CompileException` (via the pre-existing `AnalysisStratumRule` wrapper) whose `InnerException` is the same raw, uninformative `InvalidCastException` (also confirmed against unmodified `master`).

## The fix

I considered two approaches:

1. **Genuinely support a `Quantifier` in this position** — the DTD permits it, so the library arguably should handle it.
2. **Convert the crash into a clear, typed, load-time diagnostic** naming the unsupported construct and where it appeared.

I went with **(2)**. Actually supporting a variable-width quantifier as the *target* of a rewrite rule (matched-and-replaced, or matched-and-deleted) is a real semantic feature, not a mechanical fix — it changes what "the same number of Lhs and Rhs children" even means for the whole rewrite-subrule-shape dispatch (`Feature`/`Narrow`/`Epenthesis` selection) these classes are built around. That's too large and semantically risky a change to make confidently in an external repo I don't own, especially with no existing test/grammar exercising it to pin the intended behavior. A clear, fail-fast diagnostic is the conservative, reviewable option, and is a strict improvement over an opaque crash either way.

Added `PatternNodeCastExtensions` (`CastToConstraints` / `AsConstraint`) and used it at every site above instead of the unchecked cast. For any grammar built entirely from `Constraint`s — i.e. **every currently-working grammar** — behavior is byte-for-byte unchanged: the helper returns the exact same `Constraint` the cast would have produced. When a `Quantifier` (or any other non-`Constraint` `PatternNode`) shows up in one of these positions, it now raises a `CompileException` such as:

> The target (left-hand side) of a rewrite rule contains a 'Quantifier\`2', which is not supported there. Only simple segments, natural classes, and boundary markers are supported; optional or repeated segment sequences (quantifiers) are only supported within a rule's left/right environment.

## Test evidence

Full solution, before vs. after (temporarily reverted just the 6 fixed `src/` files to `master`'s content to confirm red state, then restored):

**Before the fix** (unmodified `master` + the two new tests): both new tests fail —
- `QuantifierInTargetIsRejectedWithClearError`: raw `System.InvalidCastException` thrown directly from `new Morpher(...)`, at `SynthesisRewriteRuleSpec.cs:33`.
- `QuantifierInReplacementIsRejectedWithClearError`: `CompileException` wrapping a raw `InvalidCastException` at `FeatureAnalysisRewriteRuleSpec.cs:43`.

**After the fix**: full solution build clean (0 errors, 0 new warnings), `dotnet test` on the whole solution:

```
SIL.Machine.Tokenization.SentencePiece.Tests: Passed: 4, Failed: 0
SIL.Machine.Morphology.HermitCrab.Tests:      Passed: 68, Failed: 0   (66 pre-existing + 2 new)
SIL.Machine.Translation.Thot.Tests:           Passed: 83, Failed: 0
SIL.Machine.Tests:                            Passed: 804, Failed: 0, Skipped: 3 (pre-existing, unrelated)
```

All pre-existing tests pass unchanged; only the two new tests are added.

Ran `dotnet csharpier format` on the touched files only (no wholesale reformatting).

## Test plan

- [x] Reproducing tests added, confirmed failing against unmodified `master`
- [x] Fix applied, same tests now pass
- [x] Full solution `dotnet build`: clean
- [x] Full solution `dotnet test`: all pass (962 total, 3 pre-existing unrelated skips)
- [x] `dotnet csharpier check` clean on touched files

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/464)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
@@ -50,7 +50,8 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
         {
             var newGroup = new Group<Word, ShapeNode>(name);
             foreach (
-                Constraint<Word, ShapeNode> constraint in groups[name].Children.Cast<Constraint<Word, ShapeNode>>()
+                Constraint<Word, ShapeNode> constraint in groups[name]
+                    .Children.CastToConstraints("A switch group of a metathesis rule")
             )
             {
                 Constraint<Word, ShapeNode> newConstraint = constraint.Clone();
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/EpenthesisAnalysisRewriteRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/EpenthesisAnalysisRewriteRuleSpec.cs
@@ -1,4 +1,3 @@
-using System.Linq;
 using SIL.Machine.Annotations;
 using SIL.Machine.FeatureModel;
 using SIL.Machine.Matching;
@@ -15,7 +14,11 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
             Pattern.Acceptable = IsUnapplicationNonvacuous;
             _targetCount = subrule.Rhs.Children.Count;
 
-            foreach (Constraint<Word, ShapeNode> constraint in subrule.Rhs.Children.Cast<Constraint<Word, ShapeNode>>())
+            foreach (
+                Constraint<Word, ShapeNode> constraint in subrule.Rhs.Children.CastToConstraints(
+                    "The replacement (right-hand side) of an epenthesis rewrite subrule"
+                )
+            )
             {
                 Constraint<Word, ShapeNode> newConstraint = constraint.Clone();
                 newConstraint.FeatureStruct.AddValue(HCFeatureSystem.Modified, HCFeatureSystem.Clean);
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/FeatureAnalysisRewriteRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/FeatureAnalysisRewriteRuleSpec.cs
@@ -39,8 +39,12 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
                 )
             )
             {
-                var lhsConstraint = (Constraint<Word, ShapeNode>)tuple.Item1;
-                var rhsConstraint = (Constraint<Word, ShapeNode>)tuple.Item2;
+                Constraint<Word, ShapeNode> lhsConstraint = tuple.Item1.AsConstraint(
+                    "The target (left-hand side) of a rewrite rule"
+                );
+                Constraint<Word, ShapeNode> rhsConstraint = tuple.Item2.AsConstraint(
+                    "The replacement (right-hand side) of a rewrite subrule"
+                );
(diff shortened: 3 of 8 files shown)


ddaspit (approved):
:lgtm:

> @ddaspit reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/464#-:-OyYjUwl4eIAXk-jpJWl:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #471, opened 2026-08-12
fix(hermitcrab): make metathesis switch-name order not matter

johnml1135:
## Problem

A `MetathesisRule` whose `leftSwitch` names the **earlier** of its two pattern groups throws instead of producing an analysis:

```
System.InvalidOperationException: Failed to compare two elements in the array.
 ---> System.ArgumentException: Only nodes from the same list can be compared. (Parameter 'other')
   at SIL.Machine.Annotations.ShapeNode.CompareTo(ShapeNode other)
   at SIL.Machine.Annotations.ShapeRangeFactory.Compare(ShapeNode x, ShapeNode y)
   at SIL.Machine.Annotations.Range`1.CompareTo(Range`1 other)
   at System.Linq.Enumerable.EnumerableSorter`2...
   at SIL.Machine.Morphology.HermitCrab.Morpher.Synthesize(...)
```

Both metathesis specs silently required the opposite orientation. Every pre-existing test in `MetathesisRuleTests` names the *later* group as the left switch, so the intuitive order was never exercised.

`Direction` is **not** involved — this reproduces under both `LeftToRight` and `RightToLeft`.

## Cause

`MoveNodesAfter` advances its `cur` anchor after every iterated node, whether or not that node was physically moved. In the failing orientation `beforeRightGroup` ends up equal to `leftGroup.Range.End`, so for a single-node group that node is simultaneously the move's anchor and the node being removed in the same iteration. `node.Remove()` sets its `List` to null and the following `cur.AddAfter(node)` never restores it. The orphaned node then reaches `morph.Children.OrderBy(ann => ann.Range)`, and `ShapeNode.CompareTo` throws because the two nodes belong to different lists.

Separately, `AnalysisMetathesisRuleSpec` builds its pattern in `leftSwitch`-then-`rightSwitch` order, but that pattern must match the **surface**, where the two groups appear in the opposite order from the underlying form. With the switch names reversed the analysis pattern matched nothing — so even with the crash fixed there was still no analysis.

## Fix

Both specs order the two switch groups by **pattern position** rather than by name, so a rule behaves identically whichever way its switch names are written.

For rules that already name the later group first the normalization is a **no-op**: the swap condition is false, and `AnalysisMetathesisRuleSpec` emits the same order it always did.

## Verification

Full `SIL.Machine.Morphology.HermitCrab.Tests` suite: **71/71 pass**. `dotnet csharpier check` clean.

Each half of the fix was reverted independently to confirm both are load-bearing:

| Reverted | Result |
|---|---|
| synthesis normalization | `SimpleRule_LeftSwitchNamesEarlierGroup` and its right-to-left variant reproduce the reported crash |
| analysis normalization | all three new tests fail with an empty analysis instead |
| neither | 71/71 pass |

Tests added:

- `SimpleRule_LeftSwitchNamesEarlierGroup` — differs from `SimpleRule` only in the switch-name order, and must give the same result
- `SimpleRule_LeftSwitchNamesEarlierGroup_RightToLeft` — same, with `Direction.RightToLeft`
- `ComplexRule_LeftSwitchNamesEarlierGroup` — reversed naming with a group between the two switches. Note this one passes even without the synthesis change; it guards the analysis-side ordering and the middle-group case rather than reproducing the crash.

## Scope, and what this does not close

This fixes the switch-name-order crash class. Two pre-existing hazards in the same code are left untouched and are not claimed to be fixed:

- `MoveNodesAfter` still advances `cur` past a skipped non-`Segment` node, so a switch group whose own range spans a boundary or anchor node could still anchor subsequent moves off a stale position. No current rule shape hits this.
- An unmatched switch capture has a null `Range.Start`. `beforeRightGroup`'s `.Prev` already dereferenced it before this change; the new comparison is guarded with `GroupCapture.Success` so it does not add a second such site, but the underlying assumption that both switches always match is unchanged.

## How it was found

While building generated-coverage fixtures for the HermitCrab XML surface on the `conformance-framework` branch: a fixture written to exercise `MetathesisRule@multipleApplicationOrder` wrote its switch names in the natural order and hit this. The ordering attribute turned out to be irrelevant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/471)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
@@ -31,12 +31,7 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
                 _pattern.Children.Add(node.Clone());
             }
 
-            // The analysis pattern has to match the SURFACE, where the two switch groups appear in the
-            // opposite order from the pattern that describes the underlying form. So the group that is
-            // later in the pattern is emitted first. Rules that name the later group as the left switch
-            // already satisfy that; ordering by pattern position instead of by name makes a rule behave
-            // the same whichever way round its switch names are written, and keeps this spec consistent
-            // with the synthesis spec's own normalization.
+            // This pattern matches the surface, where the switch groups appear in reverse pattern order.
             int leftIndex = Array.FindIndex(groupOrder, g => g.Name == leftGroupName);
             int rightIndex = Array.FindIndex(groupOrder, g => g.Name == rightGroupName);
             string firstName = leftIndex < rightIndex ? rightGroupName : leftGroupName;
--- a/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/PhonologicalRules/AnalysisMetathesisRuleSpec.cs
@@ -31,8 +31,18 @@ namespace SIL.Machine.Morphology.HermitCrab.PhonologicalRules
                 _pattern.Children.Add(node.Clone());
             }
 
-            AddGroup(groups, leftGroupName);
-            AddGroup(groups, rightGroupName);
+            // The analysis pattern has to match the SURFACE, where the two switch groups appear in the
+            // opposite order from the pattern that describes the underlying form. So the group that is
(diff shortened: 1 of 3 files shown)


ddaspit (approved):
:lgtm:

> @ddaspit reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/471#-:-Ozq6XXa5Lz8UNbUcGgn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #476, opened 2026-08-14
Remove deuterocanon tests

### LINKED ISSUE #174: Add deuterocanon tests to machine.py
See https://github.com/sillsdev/machine/pull/284. This needs to be ported to machine.py.

Enkidu93:
Fixes https://github.com/sillsdev/machine.py/issues/174.

I didn't see any problems in the logging from the tests, so at least at this point, I think we've resolved any issues that these would have caught. The only problems I noticed were with the tests themselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/476)


pmachapman (approved):
:lgtm:

> @pmachapman reviewed 32 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/476#-:-P-B_lAx2QeMv5wMHbqv:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Enkidu93).

___
*[`src/SIL.Machine/Corpora/ScrVersExtensions.cs` line 11 at r1](https://reviewable.io/reviews/sillsdev/machine/476#-P-BZW0-D5qdtvNy_X41:-P-BZW0-D5qdtvNy_X42:b-p0ed3c) ([raw file](https://github.com/sillsdev/machine/blob/0b58830855ddc561ee8a2b751ab948f3e9d24674/src/SIL.Machine/Corpora/ScrVersExtensions.cs#L11)):*
> ```Smalltalk
>         /// Gets a list of references (verse references) for the specified book.
>         /// </summary>
>         public static IEnumerable<VerseRef> GetReferencesForBook(this ScrVers scrVers, int bookNum)
> ```

If I understand correctly, you are removing this because it is not a part of machine.py? (as well as the fact it was only used by tests). I couldn't find any code on GitHub that use this method, but that doesn't mean that no one is necessarily using this function.



_Code quote:_
```Smalltalk
public static IEnumerable<VerseRef> GetReferencesForBook(this ScrVers scrVers, int bookNum)
```

Enkidu93:
*[`src/SIL.Machine/Corpora/ScrVersExtensions.cs` line 11 at r1](https://reviewable.io/reviews/sillsdev/machine/476#-P-BZW0-D5qdtvNy_X41:-P-EqeCi4daLa0cOfNK7:b-5t15y1) ([raw file](https://github.com/sillsdev/machine/blob/0b58830855ddc561ee8a2b751ab948f3e9d24674/src/SIL.Machine/Corpora/ScrVersExtensions.cs#L11)):*

If I understand correctly, you are removing this because it is not a part of machine.py? (as well as the fact it was only used by tests). I couldn't find any code on GitHub that use this method, but that doesn't mean that no one is necessarily using this function.
</blockquote></details>

Yes, and I feel fairly confident removing it because until recently, this class was in `SIL.Machine.Tests`, so I don't think it's being used elsewhere.

PR #479, opened 2026-08-18
add low book confidence checker

### LINKED ISSUE #477: Add a function to detect low book confidence
A function should be added that will take in a book's confidence and return True if the book confidence is below a certain threshold. The logic will be ported from machine.py so that Serval can call it.

mshannon-sil:
This replicates the logic in machine.py for checking whether a book's confidence is unusually low. It also ports the tests over.

One thing to note is it does raise an error for invalid input. I don't think this should happen, but if you think for some reason there may be scenarios where the book confidence is not between 0 and 1, Serval would need to either check for that beforehand or catch the exception, or else we can revise the function here to not throw one and return `null` instead. I'd hate for this to crash users' builds unexpectedly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/479)

### DIFF
--- a/src/SIL.Machine/QualityEstimation/BookConfidence.cs
+++ b/src/SIL.Machine/QualityEstimation/BookConfidence.cs
@@ -1,0 +1,24 @@
+﻿using System;
+
+namespace SIL.Machine.QualityEstimation
+{
+    public static class BookConfidence
+    {
+        public const double LowBookConfidenceThreshold = 0.42;
+
+        public static bool IsBookConfidenceUnusuallyLow(double confidence, string bookId = null, string model = null)
+        {
+            if (!(confidence >= 0 && confidence <= 1))
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(confidence),
+                    confidence,
+                    "The book confidence is invalid. It is calculated as the geometric mean of the segment "
+                        + "confidences, and it must be between 0 and 1, inclusive."
+                );
+            }
+
+            return confidence < LowBookConfidenceThreshold;
+        }
+    }
+}
--- a/src/SIL.Machine/QualityEstimation/BookConfidence.cs
+++ b/src/SIL.Machine/QualityEstimation/BookConfidence.cs
@@ -4,8 +4,25 @@ namespace SIL.Machine.QualityEstimation
 {
     public static class BookConfidence
     {
-        public const double LowBookConfidenceThreshold = 0.42;
+        public static readonly double LowBookConfidenceThreshold = 0.42;
 
+        /// <summary>
+        /// Determines whether a book confidence is unusually low, i.e. below
+        /// <see cref="LowBookConfidenceThreshold"/>.
+        /// </summary>
+        /// <param name="confidence">
+        /// The book confidence, calculated as the geometric mean of the segment confidences. Must be between 0 and 1,
+        /// inclusive.
+        /// </param>
+        /// <param name="bookId">The book id. Reserved for future book-specific logic; not currently used.</param>
+        /// <param name="model">The model name. Reserved for future model-specific logic; not currently used.</param>
+        /// <returns>
+        /// <c>true</c> if <paramref name="confidence"/> is below <see cref="LowBookConfidenceThreshold"/>; otherwise,
+        /// <c>false</c>.
+        /// </returns>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="confidence"/> is not between 0 and 1, inclusive.
+        /// </exception>
         public static bool IsBookConfidenceUnusuallyLow(double confidence, string bookId = null, string model = null)
         {
             if (!(confidence >= 0 && confidence <= 1))
@@ -13,8 +30,7 @@ namespace SIL.Machine.QualityEstimation
                 throw new ArgumentOutOfRangeException(
                     nameof(confidence),
                     confidence,
-                    "The book confidence is invalid. It is calculated as the geometric mean of the segment "
-                        + "confidences, and it must be between 0 and 1, inclusive."
+                    "The book confidence must be between 0 and 1, inclusive."
                 );
             }
 
--- a/tests/SIL.Machine.Tests/QualityEstimation/BookConfidenceTests.cs
+++ b/tests/SIL.Machine.Tests/QualityEstimation/BookConfidenceTests.cs
@@ -1,0 +1,59 @@
+﻿using NUnit.Framework;
+
+namespace SIL.Machine.QualityEstimation;
+
+[TestFixture]
+public class BookConfidenceTests
+{
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_AtThreshold()
+    {
+        Assert.That(BookConfidence.IsBookConfidenceUnusuallyLow(BookConfidence.LowBookConfidenceThreshold), Is.False);
+    }
+
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_MinConfidence()
+    {
+        Assert.That(BookConfidence.IsBookConfidenceUnusuallyLow(0.0), Is.True);
+    }
+
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_MaxConfidence()
+    {
+        Assert.That(BookConfidence.IsBookConfidenceUnusuallyLow(1.0), Is.False);
+    }
+
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_WithBookIdAndModel()
+    {
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(
+                BookConfidence.IsBookConfidenceUnusuallyLow(0.3, "MAT", "facebook/nllb-200-distilled-1.3B"),
+                Is.True
+            );
+            Assert.That(
+                BookConfidence.IsBookConfidenceUnusuallyLow(0.9, "MAT", "facebook/nllb-200-distilled-1.3B"),
+                Is.False
+            );
+        }
+    }
+
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_Negative()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => BookConfidence.IsBookConfidenceUnusuallyLow(-0.5));
+    }
+
+    [Test]
+    public void IsBookConfidenceUnusuallyLow_GreaterThanOne()
+    {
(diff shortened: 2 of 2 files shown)


pmachapman (changes requested):
> @pmachapman reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/479)* status: all files reviewed, 2 unresolved discussions (waiting on Enkidu93 and mshannon-sil).

___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-ITEgd-xRFDOW8K71v:-P-ITEgd-xRFDOW8K71w:b3lhfcl) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L7)):*
> ```Smalltalk
>     public static class BookConfidence
>     {
>         public const double LowBookConfidenceThreshold = 0.42;
> ```

Can you please make this a `static readonly` instead of a `const`, as there is a chance it may change in future?

```cs
public static readonly double LowBookConfidenceThreshold = 0.42;
```

(A const is evaluated at compile time, a static readonly is evaluated at runtime. Making this a const could cause inconsistencies between assemblies that reference this value and the `IsBookConfidenceUnusuallyLow` function below if they update their reference to the library but are not themselves recompiled.)



_Code quote:_
```Smalltalk
public const double LowBookConfidenceThreshold = 0.42;
```
___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-IUl9_EhOVtKQkA_RR:b-srk2n5) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*
> ```Smalltalk
>         public const double LowBookConfidenceThreshold = 0.42;
> 
>         public static bool IsBookConfidenceUnusuallyLow(double confidence, string bookId = null, string model = null)
> ```

Can you please create a codedoc comment that documents the behavior of this function, as I think the `ArgumentOutOfRangeException` may be non-obvious to someone implementing a call to this function in the future?

```cs
/// <summary>
/// 
/// </summary>
/// <param name="confidence"></param>
/// <param name="bookId"></param>
/// <param name="model"></param>
/// <returns></returns>
/// <exception cref="ArgumentOutOfRangeException"></exception>
```



_Code quote:_
```Smalltalk
public static bool IsBookConfidenceUnusuallyLow(double confidence, string bookId = null, string model = null)
```

Enkidu93 (commented):
> @Enkidu93 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/479)* status: all files reviewed, 2 unresolved discussions (waiting on mshannon-sil and pmachapman).

___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-KWQI78Uh0pUcpBwlA:b-365uzp) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*

Can you please create a codedoc comment that documents the behavior of this function, as I think the `ArgumentOutOfRangeException` may be non-obvious to someone implementing a call to this function in the future?

```cs
/// <summary>
/// 
/// </summary>
/// <param name="confidence"></param>
/// <param name="bookId"></param>
/// <param name="model"></param>
/// <returns></returns>
/// <exception cref="ArgumentOutOfRangeException"></exception>
```
</blockquote></details>

Yeah, I agree. It would be reasonable for the geometric mean function to throw an error like this, but it feels a little out-of-place in this function. We could throw an exception as needed in Serval while calculating the geometric mean. I'm fine with either leaving this or just removing the exception-throwing altogether. If we do keep this exception, maybe change 'It is...' to 'It should be...'. I'm not sure: This exception should be about this function not making sense for values outside of 0-1, not about how the geometric mean requires non-negative values. This all makes me think that the `QualityEstimation` code itself should be calculating the mean :thinking:.

mshannon-sil:
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-L-hHz06H6nJbMdJTu:b-x5542s) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*

Yeah, I agree. It would be reasonable for the geometric mean function to throw an error like this, but it feels a little out-of-place in this function. We could throw an exception as needed in Serval while calculating the geometric mean. I'm fine with either leaving this or just removing the exception-throwing altogether. If we do keep this exception, maybe change 'It is...' to 'It should be...'. I'm not sure: This exception should be about this function not making sense for values outside of 0-1, not about how the geometric mean requires non-negative values. This all makes me think that the `QualityEstimation` code itself should be calculating the mean :thinking:.
</blockquote></details>

Yeah let me change the exception text to not mention the geometric mean, and just focus on it needing to be between 0 and 1.

When you say you think that the quality estimation code itself should be calculating the mean, can you clarify what you're thinking there? Is the suggested workflow that Serval might pass the segment-level confidence to a method in Machine's QualityEstimation, which would calculate the book confidence and also return the boolean?

mshannon-sil (commented):
> @mshannon-sil made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/479)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on Enkidu93 and pmachapman).

___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-ITEgd-xRFDOW8K71v:-P-L20HQ6vq4eXk8ZyGF:b-896fix) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L7)):*

Can you please make this a `static readonly` instead of a `const`, as there is a chance it may change in future?

```cs
public static readonly double LowBookConfidenceThreshold = 0.42;
```

(A const is evaluated at compile time, a static readonly is evaluated at runtime. Making this a const could cause inconsistencies between assemblies that reference this value and the `IsBookConfidenceUnusuallyLow` function below if they update their reference to the library but are not themselves recompiled.)
</blockquote></details>

Done.
___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-L2--4EZTXON2FK7my:b-ywbop9) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*
<details><summary><i>Previously, mshannon-sil wrote…</i></summary><blockquote>

Yeah let me change the exception text to not mention the geometric mean, and just focus on it needing to be between 0 and 1.

When you say you think that the quality estimation code itself should be calculating the mean, can you clarify what you're thinking there? Is the suggested workflow that Serval might pass the segment-level confidence to a method in Machine's QualityEstimation, which would calculate the book confidence and also return the boolean?
</blockquote></details>

I simplified the exception message substantially. If you still think that we shouldn't raise an exception at all, I can remove it.

pmachapman (approved):
:lgtm:

> @pmachapman reviewed 1 file and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/479#-:-P-LHFs_DTrLmnSQ5XLR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

Enkidu93 (approved):
:lgtm:

> @Enkidu93 reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/479#-:-P-LUuav6iP4vHncZhda:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

___
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-LUS2HEJY3FfFSnBCQ:b-2z9jqc) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*
<details><summary><i>Previously, mshannon-sil wrote…</i></summary><blockquote>

I simplified the exception message substantially. If you still think that we shouldn't raise an exception at all, I can remove it.
</blockquote></details>

Great, thank you!

Well, it seems like if we want silnlp and Serval to be aligned more fully, it would make sense to move this whole process to a shared utility since there are a number of different steps at which they could diverge (e.g., which segments are filtered out, how the mean is taken, etc.). For now, this is good, but I think we should consider refactoring it in the future.

mshannon-sil:
*[`src/SIL.Machine/QualityEstimation/BookConfidence.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/machine/479#-P-IUl9_EhOVtKQkA_RQ:-P-LXARv69WPCEg3qEj2:bx22v0b) ([raw file](https://github.com/sillsdev/machine/blob/3d6eee2cc3ea7a80bf029657ad784b7aa6a3f279/src/SIL.Machine/QualityEstimation/BookConfidence.cs#L9)):*

Great, thank you!

Well, it seems like if we want silnlp and Serval to be aligned more fully, it would make sense to move this whole process to a shared utility since there are a number of different steps at which they could diverge (e.g., which segments are filtered out, how the mean is taken, etc.). For now, this is good, but I think we should consider refactoring it in the future.
</blockquote></details>

Yeah, that makes sense!

PR #489, opened 2026-08-27
Fold-step synthesis memo: built, correct, and does not pay (evidence, not for merge)

johnml1135:
Stacked on #488. **This is the evidence that closes a family of optimizations. It is not proposed as a performance feature.**

#488 measured 3.22x and 8.10x shareable synthesis fold steps, with ~50%% wall-clock ceilings on the two conformance fixtures large enough to time reliably, reproduced across two independent runs. This branch builds it.

**It is correct.** `SynthesisStateKey` carries the ordered remaining trail plus shape, syntactic FS, realizational FS, MPR set, root allomorph, disjunctive allomorph indices, applied counts, `IsPartial`, `IsLastAppliedRuleFinal` and stratum — each field justified in the doc comment against the code that reads it. Stored outputs are re-anchored rather than handed over with a foreign trail embedded. Steps are set-valued, because realizational rules are trail-exempt. Off by default. **Parity: 0 divergences across all 33 conformance fixtures.**

**It does not pay.** Warm-up discarded, min of 5 interleaved samples per arm:

| fixture | predicted | realised | memo hits | off-arm noise |
| --- | --- | --- | --- | --- |
| deep-optional-affix-nesting | 50.4%% (~2x) | **0.96x** (4.5%% slower) | **0** | 38.6%% |
| suffixing-evidential-adjacency-chain | 52.0%% (~2x) | 1.06x | 2,682 | 21.9%% |

`hits = 0` on the fixture with the largest reliable sample and the highest ceiling. With a trail-complete key the memo never fires there; the regression is key-construction cost with no payoff.

**Why, and why it generalises.** A sound key must carry the remaining trail, so two candidates have to agree on their entire future to share a step. That is the third independent route to the same result:

| measurement | key | apparent | sound |
| --- | --- | --- | --- |
| synthesis-input dedupe | order-insensitive | 9,774x | 15-40%% |
| fold-entry census | trail position only | 6,476x | not established |
| fold-step sharing (here) | trail position only | 3.22x / 8.10x | **hits = 0** |

**The redundancy is apparent, not real. The trail is what makes each step distinct, and every measurement showing large shareable work is measuring a key that omits it.** Packed parse forests, fold-step sharing and synthesis-input dedupe all require distinct derivations to converge on a genuinely identical state. In this engine they do not converge — the same fact as the rules being non-order-invariant, seen from the other side.

**Disposition.** Recorded as row 5 in `docs/hermitcrab-optimization-ledger.md`. Merge only if you want the implementation in tree as executable proof; otherwise close and keep the branch. Either way the ledger row is the durable artifact, and it lives in #488.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/489)

### DIFF
--- a/docs/hermitcrab-synthesis-fold-probes.md
+++ b/docs/hermitcrab-synthesis-fold-probes.md
@@ -626,3 +626,52 @@ Commit `1a7d484c` bundles the harness reporting change with the section 8 analys
 agent that wrote the reporting change stopped with it uncommitted and the work was swept in when
 the analysis was committed. Two concerns in one commit; flagged here rather than rewritten, since
 the PR body is the review artifact.
+---
+
+## 10. The build was made, and it does not pay — because the redundancy is not real
+
+`feature/synthesis-fold-sharing` (`1048c742`) implements the fold-step memo with a sound key:
+`SynthesisStateKey` carries the **ordered remaining trail**, not just its index, plus shape,
+syntactic FS, realizational FS, MPR set, root allomorph, disjunctive allomorph indices, applied
+counts, `IsPartial`, `IsLastAppliedRuleFinal` and stratum — each field justified against its reader.
+
+**Parity: 0 divergences across all 33 conformance fixtures.** The implementation is correct.
+
+**Measured (warm-up discarded, min of 5 interleaved samples per arm):**
+
+| fixture | predicted ceiling | realised | memo hits | off-arm spread |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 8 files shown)


johnml1135:
Superseded by #490 (docs-only). The implementation is preserved on `feature/synthesis-fold-sharing` as reproducible evidence — parity-clean across all 33 conformance fixtures, `hits = 0` on the best case — and is row 5 of the ledger. Not merging research scaffolding into product assemblies.

PR #494, opened 2026-09-04
Change Add to PriorityUnion

jtmaxwell3:
I fixed a performance bug by changing AnalysisAffixProcessRule.Apply to use PriorityUnion instead of Add.  Add acts like Union, PriorityUnion is more restrictive.  Also, SynthesisAffixProcessRule uses PriorityUnion instead of Add in the other direction, so this makes AnalysisAffixProcessRule more parallel to SynthesisAffixProcessRule.

If I have a sequence of derivation rules that change the category like [CAT:A] => [CAT:B], [CAT:B] => [CAT:A], followed by [CAT:B] => [CAT:C], then the syntactic feature structure of the input of the third rule will be [CAT:{A, B}] if I use Add, and the third rule will not be filtered when it checks whether the input can unify with [CAT:B].  If I use PriorityUnion, then the syntactic feature structure of the input of the third rule will be [CAT:A], and the third rule will be filtered when it checks whether the input can unify with [CAT:B].

This performance bug only shows up when there are at least three derivational affixes.  But words that are slow to parse try many long sequences of derivational affixes.  Filtering these long sequences can make a big difference.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/494)

### DIFF
--- a/src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs
+++ b/src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs
@@ -56,7 +56,7 @@ namespace SIL.Machine.Morphology.HermitCrab.MorphologicalRules
                 foreach (Word outWord in _rules[i].Apply(input).RemoveDuplicates())
                 {
                     if (!_rule.RequiredSyntacticFeatureStruct.IsEmpty)
-                        outWord.SyntacticFeatureStruct.Add(_rule.RequiredSyntacticFeatureStruct);
+                        outWord.SyntacticFeatureStruct.PriorityUnion(_rule.RequiredSyntacticFeatureStruct);
                     else if (_rule.OutSyntacticFeatureStruct.IsEmpty)
                         outWord.SyntacticFeatureStruct.Clear();
                     outWord.MorphologicalRuleUnapplied(_rule);


ddaspit (changes requested):
It would be great if you could add a unit test to capture scenario you laid out.

> @ddaspit reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/494#-:-P0hug2Z8NBnTGuYGdnp:b-o3nt8g)* status: all files reviewed, 1 unresolved discussion (waiting on jtmaxwell3).

___
*[`src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs` line 59 at r1](https://reviewable.io/reviews/sillsdev/machine/494#-P0humxUDsky3DNjIumn:-P0humxUDsky3DNjIumo:bb908jc) ([raw file](https://github.com/sillsdev/machine/blob/dc8efec51ec34f34c5c2e1a04bf9b1d79a471a5d/src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs#L59)):*
> ```Smalltalk
>                 {
>                     if (!_rule.RequiredSyntacticFeatureStruct.IsEmpty)
>                         outWord.SyntacticFeatureStruct.PriorityUnion(_rule.RequiredSyntacticFeatureStruct);
> ```

This change should also apply to compounding rules as well.

johnml1135 (commented):
I measured this on four FieldWorks grammars, 30 words each, counting the parser's steps rather than timing it, and comparing every parse by gloss, allomorph, and features.

| Grammar | Feature checks | Synthesis attempts | Wall time |
| --- | ---: | ---: | ---: |
| Mbugwe | 14.7M to 4.0M | 11.5M to 5.4M | 340 s to 133 s |
| Sena | 747k to 138k | 340k to 70k | 28 s to 8.5 s |

Amharic and Indonesian: no change. Every word got exactly the same parses. All 36 conformance grammars pass, and 1,300 words generated from them parse the same as master.

**One regression, with a fix.** The parser merges two partial analyses whenever they have the same shape, even if they carry different feature requirements (`MergeEquivalentAnalyses` in `AnalysisStratumRule`). It then keeps working on only one of them and copies the results back to the other at the end. With the old Add, the kept one always had the looser requirements, so nothing was lost. With PriorityUnion, the kept one can be the stricter one, and a rule that was valid for the other analysis gets filtered out. A real parse disappears. The fix: when merging, loosen the kept analysis's features to cover both (`FeatureStruct.Union`). On Mbugwe this fires about 100k times with no change to any parse and under 1 percent change in step counts. Test included, fails without the fix.

Branch `perf/pr494-proposed` (one commit, 3be1eb96, on top of yours) has this fix, the same PriorityUnion change in `AnalysisCompoundingRule` as Damien asked, and three tests including the three-rule chain from your description. Happy to open it against this branch or you can cherry-pick.

src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs
```diff
@@ -56,7 +56,7 @@ public IEnumerable<Word> Apply(Word input)
                 foreach (Word outWord in _rules[i].Apply(input).RemoveDuplicates())
                 {
                     if (!_rule.RequiredSyntacticFeatureStruct.IsEmpty)
-                        outWord.SyntacticFeatureStruct.Add(_rule.RequiredSyntacticFeatureStruct);
+                        outWord.SyntacticFeatureStruct.PriorityUnion(_rule.RequiredSyntacticFeatureStruct);
```
johnml1135, line 59: Agreed, this is the right fix. When the parser builds a word (synthesis), it asks "does this stem match what the rule requires?" So when it takes a word apart (analysis), the stem left behind must match exactly what the rule requires. That is what PriorityUnion says. The old Add kept a running "any of these categories" list, which let obviously wrong stems through to be rejected later.

I tried hard to break this: 21 targeted unit tests, all 36 conformance grammars, and 1,300 words generated from those grammars, parsed on master and on this branch. Every word got the same parses. The one place it does bite is described in my general comment.

src/SIL.Machine.Morphology.HermitCrab/MorphologicalRules/AnalysisAffixProcessRule.cs
```diff
@@ -56,7 +56,7 @@ public IEnumerable<Word> Apply(Word input)
...
                     else if (_rule.OutSyntacticFeatureStruct.IsEmpty)
                         outWord.SyntacticFeatureStruct.Clear();
```
johnml1135, line 61: Not for this PR, but while we are here. When a rule has no required features, the features its output set are left sitting on the stem. So if an inner suffix marks "present" and an outer suffix marks "past", the parser sees "past" on the stem, then refuses to take off the inner suffix because it expects "present". That parse is lost today, on master and on this branch alike. Wiping the rule's output features off the stem here would find it, and in my measurements it also cuts another 30 to 45 percent of the work on Mbugwe and Sena. It needs one other change first, so: follow-up.

jtmaxwell3:
@johnml1135: There is a bug in MergeEquivalentAnalyses that I attempt to fix in https://github.com/sillsdev/machine/pull/493.  The fix is based on AnalysisStateKey.  I think that this is cleaner than using FeatureStruct.Union.

jtmaxwell3:
@ddaspit: I prefer the fix to MergeEquivalentAnalyses in https://github.com/sillsdev/machine/pull/493, but I defer to you.

ddaspit (commented):
I would like to see two things:

1. Add a unit test
2. Make the same change for the compounding rule

> @ddaspit made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/494#-:-P1G2Cc31xHxA54zt6o-:bstj6qy)* status: all files reviewed, 3 unresolved discussions (waiting on jtmaxwell3).

johnml1135:
I found a recall regression in this change. A word that master analyses correctly gets no analysis under PriorityUnion. I reproduced it on your branch's own build, so it is not a porting artifact.

### The grammar

One root `zud` and four suffix rules on a three-symbol `num` feature:

| rule | Required | Out | suffix |
| --- | --- | --- | --- |
| A | `num=du` | `num=pl` | `i` |
| B | *(none)* | `num=sg` | `u` |
| C | `num=sg` | `num=pl` | `e` |
| D | `num=pl` | *(none)* | `o` |

Rule B is the shape that matters: an output feature and no required feature.

### The result

Built from `d3b7643d` (master semantics) and from `2acb3c52` (this branch), same grammar, same word list:

| word | morphemes | master (Add) | this branch (PriorityUnion) |
| --- | --- | --- | --- |
| `zud` | ZUD | `ZUD` | `ZUD` |
| `zudi` | +A | `ZUD+A` | `ZUD+A` |
| `zudiu` | +A+B | `ZUD+A+B` | `ZUD+A+B` |
| `zudiue` | +A+B+C | **no analysis** | **no analysis** |
| `zudueo` | +B+C+D | `ZUD+B+C+D` | `ZUD+B+C+D` |
| `zudiueo` | +A+B+C+D | `ZUD+A+B+C+D` | **no analysis** |

`zudiueo` is the regression. It is a well-formed word: `zud` → A gives `pl` → B gives `sg` → C requires `sg` and gives `pl` → D requires `pl`. Master finds it. This branch does not.

### Why

Analysis carries an accumulated feature structure that is meant to describe the stem at the current point. Un-applying B is where it goes wrong. B declares an output feature and no required feature, and neither Add nor PriorityUnion removes B's own output from the accumulated structure. So `sg`, which B *wrote*, is carried backwards as if it *constrained* B's input. The stem entering B was really `pl`, from A's output. A's gate then checks its output `pl` against a stem the fold claims is `sg`, and refuses.

That single mechanism predicts every cell in the table:

- Under Add, `zudiueo` survives by luck. D puts `pl` on the accumulated structure first, and C's Add unions rather than replaces, leaving `{sg, pl}`, which still overlaps A's `pl`.
- `zudiue` has no D, so the accumulated structure is `sg` alone, and Add loses it too. **Master already loses this word.**
- Under PriorityUnion, C's required `sg` overwrites D's `pl`, so even `zudiueo` is lost.

So PriorityUnion does not introduce the underlying bug, it removes the accident that was hiding it. Add's widening was masking a stale output feature.

### The fix

This is exactly what the follow-up in https://github.com/sillsdev/machine/pull/494#discussion_r3974223557 anticipated: wiping the rule's output features off the stem. Doing that makes the accumulated structure mean what analysis needs it to mean, and it fixes both words rather than only the regression.

Concretely, the analysis fold becomes the exact inverse of synthesis. Synthesis computes `PriorityUnion(Unify(stem, required), out)`. So analysis should gate on `PriorityUnion(required, out)` (strictly stronger than the current `out` check, still necessary), strip the feature paths that `out` defines off the stem, then unify with `required`, and never clear the structure.

I have that implemented in the Rust port of HermitCrab and measured it. On the six words above it agrees with master everywhere master is right, and additionally finds `zudiue`, which both folds currently lose. Across five real FieldWorks grammars it changes no analysis set at all, and it is faster than PriorityUnion: on 30-word samples, rule attempts drop 1.3x on Mbugwe, 1.6x on Amharic, 3.6x on Amharic's pinned worst words, with Sena flat.

I am happy to put the grammar files and both TSV outputs somewhere you can run them, or to open a PR against this branch with the exact-inverse fold and these cases as tests. One caveat worth stating: the deeper change needs feature-structure-aware deduplication in the template battery first, since `AnalysisAffixTemplateRule` collapses two templates that share a slot rule but differ in required category, which is the "one other change" the earlier comment referred to.

johnml1135:
Follow-up with the wider check I offered, and one correction to my comment above.

I widened the sweep to 11 grammars, each run through four builds: master (`d3b7643d`), this branch (`2acb3c52`), and two builds of the Rust port. The new grammars vary the shape deliberately: a two-symbol feature so `AddImpl`'s delete-on-cover branch is hit, a five-rule chain with two consecutive no-required rules, a nested `ComplexFeature` with an inert sibling subfeature, a `CompoundingRule` exercising `HeadRequiredSyntacticFeatureStruct` rather than a fourth affix, the same recipe declared as `AffixTemplate` slots, `multipleApplication="2"` so one rule genuinely fires twice, a second independent symbol permutation, and a control where no rule's gate is ever checked against a value written more than one layer out.

**The correction: master under-generates in this class too, not just this branch.**

My first comment framed `zudiueo` as a regression against a master that was right. That is true for that word, but incomplete. In 7 of the 8 new grammars there is a *shorter* word, the same chain with the outermost rule omitted, that master also loses:

| grammar | word master loses | why it is well-formed |
| --- | --- | --- |
| g04 | `dofpkt` | A+B+C, D skipped |
| g05 | `feniuxe` | A+B+B2+C, D skipped |
| g06 | `cotaou` | A+B+C, D skipped |
| g07 | `bimtokue` | compound + B+C, D skipped |
| g08 | `wugaue` | templated A+B+C, D skipped |
| g09 | `wigauue` | B applied twice, D skipped |
| g11 | `hoxpqr` | A+B+C, D skipped |

Each was hand-traced forward: root, then each rule's required features checked against the immediately preceding rule's output, through to the surface. They synthesize. Master's `Add` loses them for the same stale-output-feature reason described above; it only rescues the full chain because the outermost rule happens to put the needed value on the accumulated structure before the union runs. Remove that rule and the rescue disappears.

The control grammar shows all four builds agreeing everywhere, so this is not an artifact of how I am running them.

Across all 11 grammars the exact-inverse fold never lost a parse that any other build found, and it found every word in the table above. That is the case for treating this as one bug with one fix rather than a regression to patch.

**Separately, a signature-rendering bug.** While tracing one of these I found that `BuildSignature` can drop a morpheme ID: when a zero-width identity rule (`CopyFromInput` with no `InsertSegments`) wraps immediately outside a rule whose own affix ends at the same shape-node boundary, the inner rule's ID is missing from the output string. In my g1 grammar, master reports `SAG+A2B+THIRD` for a derivation that actually applied `a2b`, `b2a`, then `third`. The analyses themselves are correct and agree with the Rust port; only the rendering differs. It is unrelated to this PR and I am happy to file it separately if useful.

I can push the 11 grammars, word lists and all four TSV outputs to a branch so you can rerun them, and I am glad to open a PR with the exact-inverse fold and these words as test cases, against this branch or against master, whichever you prefer.

jtmaxwell3:
@johnml1135 The merge with master that occurred at the same time may have fixed this, since it includes related fixes from https://github.com/sillsdev/machine/commit/52d069f845b43f8bc88a95a56a8511aa58def26f.  Could you please run your checks again?  I also added the unit tests from perf/pr494-proposed, which failed before the merge but succeeded after it.

johnml1135:
The adversarial tests I wrote while reviewing this change are pushed on the branch `perf/pr494-priority-union`, if they're useful to you:

https://github.com/sillsdev/machine/tree/perf/pr494-priority-union

The file is `tests/SIL.Machine.Morphology.HermitCrab.Tests/MorphologicalRules/AnalysisSyntacticFeatureMergeTests.cs`. They cover the merge-loses-a-parse behaviour. The case to look at is `ShapeMerge_OneRulePathFirst_NoWidening_PriorityUnionLosesParse`: `Morpher.MergeEquivalentAnalyses` folds same-shape stratum outputs into a single canonical word using the FS-blind `Word.ValueEquals`, so when the canonical word's stem constraint is narrower than an alternate path's, a lower-stratum rule that is valid for that other path gets filtered out and a well-formed parse is lost. Under `Add` that test finds 2 analyses; under `PriorityUnion` it finds 0.

One caveat: that is a research branch rather than a merge candidate. The tests run against a small harness (`AnalysisSyntacticFeatureMerge`) that lets one binary switch between `Add`, `PriorityUnion` and `Exact` through the `HC_ANALYSIS_FS_MERGE` environment variable, so the three modes can be compared side by side. That harness is threaded through several production files there, which is why the test file doesn't drop cleanly onto this PR's branch on its own. Background notes are in `docs/pr494-review.md` on the same branch.

jtmaxwell3:
@johnml1135: Thanks for the adversarial tests!  Merging master into it made the problem go away.

PR #497, opened 2026-09-14
Add TryGetChapters to ScriptureRangeParser

### LINKED ISSUE #486: Add a TryGetChapters function to ScriptureRangeParser
When I call `ScriptureRangeParser.GetChapters()`, I sometimes need to wrap it in a try/catch to ensure that any invalid chapter ranges (i.e. because a versification has changed) do not crash the method calling it, and I would rather the method just skip the range.

Does it make sense to have a `ScriptureRangeParser.TryGetChapters(string scriptureRange, out Dictionary<string, List<int>> chapters)` method that returns `true` if the chapters were retrieved, or `false` if they were not?

Another option could be to have a more forgiving GetChapters() implementation that will just parse the chapter ranges specified, even if they are not precisely the same as the versification requires (i.e. additional chapter numbers are in the range)

In Scripture Forge I created an extension method along the lines of:

```cs
    public static bool TryGetChapters(
        this ScriptureRangeParser scriptureRangeParser,
        string chapterSelections,
        [NotNullWhen(true)] out Dictionary<string, List<int>>? chapters
    )
    {
        try
        {
            chapters = scriptureRangeParser.GetChapters(chapterSelections);
            return true;
        }
        catch (ArgumentException)
        {
            chapters = null;
            return false;
        }
    }
```

pmachapman:
Fixes #486 

I noticed that the ScriptureRangeParser was not in a namespace, so set the namespace to SIL.Machine.Scripture. If it was not an oversight, I can revert it to remain in the SIL.Machine namespace.

I also took the opportunity to update the unit tests to use new .Net 10 language features.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/machine/497)


Enkidu93 (approved):
:lgtm:

> @Enkidu93 reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/497#-:-P1WKhT_3kbf7c5rZn1W:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ddaspit).

src/SIL.Machine/Scripture/ScriptureRangeParser.cs
```diff
@@ -5,244 +5,280 @@
...
+        }
+
+        public bool TryGetChapters(string chapterSelections, out Dictionary<string, List<int>> chapters)
+        {
+            try
+            {
+                chapters = GetChapters(chapterSelections);
+                return true;
+            }
+            catch (ArgumentException)
```
ddaspit, line 277: **Critical** — catching only `ArgumentException` lets an empty section escape as an unhandled `InvalidOperationException` from `section.Last()` (line 53), so `TryGetChapters` throws instead of returning `false`.

Confirmed on a local build of this branch: `"MAT;"`, `"MAT;;MRK"`, `";"`, and `"  ;  "` all throw. That is the input class the Try* overload exists to absorb.

Either skip empty sections in `GetChapters`, or throw `ArgumentException` for them so this catch covers them.

src/SIL.Machine/Scripture/ScriptureRangeParser.cs
```diff
@@ -5,244 +5,280 @@
...
+        public static Dictionary<string, List<int>> GetChapters(
+            string chapterSelections,
+            ScrVers versification = null
+        ) => new ScriptureRangeParser(versification).GetChapters(chapterSelections);
 
-        //*Specific chapters from one book*
-        if (char.IsDigit(section.Last()) && section.Length > 3)
+        public static bool TryGetChapters(
+            string chapterSelections,
+            ScrVers versification,
```
ddaspit, line 33: **Minor** — `versification` sits before the `out` parameter, so it cannot carry the `ScrVers versification = null` default that static `GetChapters` has just above; every caller must pass `null` explicitly.

Moving it after `out chapters`, or adding a two-argument overload, restores parity with `GetChapters`.

ddaspit (commented):
Two findings from a local review: one blocking, one minor. Built the branch (0 warnings) and ran the 86 existing `ScriptureRangeParser` tests (all pass); the empty-section behavior below was confirmed with a throwaway test against this branch.

src/SIL.Machine/Scripture/ScriptureRangeParser.cs
pmachapman: Done.

src/SIL.Machine/Scripture/ScriptureRangeParser.cs
pmachapman: Done. I updated the tests to accept an exception type to support this.

pmachapman (commented):
> @pmachapman made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/497)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit).

src/SIL.Machine/Scripture/ScriptureRangeParser.cs
ddaspit: I would prefer to create an overload rather than move it after the `out` parameter.

ddaspit (commented):
> @ddaspit reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/machine/497)* status: all files reviewed, 1 unresolved discussion (waiting on pmachapman).

